PR #15, opened 2024-09-09
feat: implement redis queue

lotharking:


src/controllers/message/MessageController.ts
```diff
@@ -75,284 +49,24 @@ export class MessageController {
...
     try {
-      const agent = await this.agentService.getAgent()
-
-      let messageId: string | undefined
-      const messageType = message.type
-      this.logger.debug!(`Message submitted. ${JSON.stringify(message)}`)
-
-      const connection = await agent.connections.findById(message.connectionId)
+      let connection: ConnectionRecord | null = null
+      connection = await this.checkForDuplicateId(message)
```
genaris, line 53: IMHO this method should return undefined, throwing an Error in case of duplicate like it does right now), and the line `const connection = await agent.connections.findById(message.connectionId)` remain here.

src/controllers/message/services/MessageServiceFactory.ts
```diff
@@ -0,0 +1,22 @@
...
+import { CoreMessageService } from './CoreMessageService'
+
+@Injectable()
+export class MessageServiceFactory {
+  constructor(
+    @Optional() @InjectQueue('message') private messageQueue: Queue,
+    private readonly coreMessageService: CoreMessageService,
+  ) {}
+
+  async setProcessMessage(redisAvailable: boolean, message: IBaseMessage, connection: ConnectionRecord) {
```
genaris, line 17: Why don't call this simply `processMessage`, receiving the message and connection and doing the isRedisAvailable check right here?

src/controllers/message/services/MessageServiceFactory.ts
```diff
@@ -0,0 +1,22 @@
...
+    return redisAvailable
+      ? await this.messageQueue.add('', { message, connection })
```
genaris, line 19: What does that '' mean? Seems a bit awkward to me.

README.md
```diff
@@ -34,6 +34,8 @@ but likely needed for production and test deployments.
 | POSTGRES_PASSWORD         | PosgreSQL database password                                                                                         | None                            |
 | POSTGRES_ADMIN_USER       | PosgreSQL database admin user                                                                                       | None                            |
 | POSTGRES_ADMIN_PASSWORD   | PosgreSQL database admin password                                                                                   | None                            |
+| REDIS_HOST       | Redis database host user                                                                                       | None                            |
```
genaris, line 37: Add some comments regarding the usage of this Redis (what does it does) and the fact that is optional.

genaris (changes requested):
Seems very interesting!

src/controllers/message/services/MessageServiceFactory.ts
lotharking: its a [conditional ternary operator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Conditional_operator). It's used to simplify the "if" condition

README.md
lotharking: fixed

src/controllers/message/services/MessageServiceFactory.ts
lotharking: fixed

src/controllers/message/MessageController.ts
lotharking: fixed

src/controllers/message/services/MessageServiceFactory.ts
genaris: I'm asking about `await this.messageQueue.add('', { message, connection })`. The first argument is an empty string. What does it mean? Is it the right way of doing it?

lotharking: You're right, it's not necessary

genaris (approved):

PR #20, opened 2024-09-25
chore: update chatbot demo with features for didcomm calls

genaris:


lotharking (changes requested):
IMHO the file phone-cred-def-dev.json must be in the docker file

lotharking (approved):

PR #31, opened 2024-10-22
feat: processed parsed data

lotharking:


src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
```diff
@@ -17,13 +59,77 @@ export class EMrtdDataSubmitMessage extends BaseMessage {
...
+        default:
+          throw new Error(`Unrecognized MRZ length: ${len}`)
+      }
+    })(mrzString.length)
+    const parsedMrz = Mrz.parse(formattedMrz)
+
+    const newEmrtdData: EMrtdRawData = {
+      raw: raw,
+      processed: {
+        ef_dg1: {
```
genaris, line 102: There is no need to put fields within specific objects for each EF. It is simpler for the clients to have a plain object containing all processed, intepreted fields without the need of dealing with something that is very specific to ICAO 9303 spec.

This means that the structure of the resulting object would be:
```
{
raw: { COM: 'xxx', DG1: 'yyy', ... },
processed: {
      documentType: parsedMrz.format,
          documentNumber: parsedMrz.documentNumber,
          issuingState: parsedMrz.fields.issuingState,
          dateOfExpiry: parsedMrz.fields.expirationDate,  // TODO: Check and specify date format
          sex: parsedMrz.fields.sex,
          nationality: parsedMrz.fields.nationality,
          lastName: parsedMrz.fields.lastName,
          firstName: parsedMrz.fields.firstName,
          mrzOptionalData: parsedMrz.fields.optional1,
          faceImages: dataUrls,
          nameOfHolder: parsed.fields.additionalPersonalData?.nameOfHolder ?? `${parsedMrz.fields.lastName} ${firstName: parsedMrz.fields.firstName}`,
          dateOfBirth: parsed.fields.additionalPersonalData?.fullDateOfBirth ?? parsedMrz.fields.birthDate, // TODO: Check and specify date format
          otherNames: parsed.fields.additionalPersonalData?.otherNames,
          personalNumber: parsed.fields.additionalPersonalData?.personalNumber,
          placeOfBirth: parsed.fields.additionalPersonalData?.placeOfBirth,
          permanentAddress: parsed.fields.additionalPersonalData?.permanentAddress,
          telephone: parsed.fields.additionalPersonalData?.telephone,
          profession: parsed.fields.additionalPersonalData?.profession,
          title: parsed.fields.additionalPersonalData?.title,
          personalSummary: parsed.fields.additionalPersonalData?.personalSummary,
          custodyInformation: parsed.fields.additionalPersonalData?.custodyInformation,
}
```

src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
```diff
@@ -1,9 +1,51 @@
...
+  processed: {
+    ef_dg1: EF_DG1
+    ef_dg2: EF_DG2
+    ef_dg11?: EF_DG11
+  }
+}
+
+export type EF_DG1 = {
+  documentType: string
+  documentNumber: string | null
```
genaris, line 19: I don't think this can be null

src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
```diff
@@ -1,9 +1,51 @@
...
+  issuingState?: string | null
+  dateOfBirth?: string | null
+  dateOfExpiry?: string | null
+  sex?: string | null
+  nationality?: string | null
+  lastName?: string | null
+  firstName?: string | null
+  nameOfHolder?: string
+  mrzOptionalData?: string | null
```
genaris, line 28: Most of these fields are mandatory, so I don't think it should be allowed them to be undefined.

BTW, this way of defining the fields will result in 3 different possibilities: undefined, string and null. I think you should choose between either returning a null when a particular, optional field (such as `mrzOptionaldata`) is not present or simply not defining it. I mean:
- `mrzOptionalData?: string` or
- `mrzOptionalData: string | null`

I'd say the first one is simpler to handle, but I don't have a strong opinion yet.

genaris (changes requested):

src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
lotharking: fixed, but I'm still unsure about the format to use. The passports maintain the YYMMDD format, so for now, I believe that should be fine

src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
lotharking: fixed

src/model/messages/mrtd/EMrtdDataSubmitMessage.ts
lotharking: fixed, I would prefer to leave it as `mrzOptionalData?: string`

genaris (approved):

PR #41, opened 2024-11-15
Feat/creation nestjs-client for chatbot deployment and example chatbot demo

lotharking:
**Features:**  
- The `nestjs-client` module was implemented to be used as a library in any project.  
- Modules for handling messages and connections were added to the new library.  
The library introduced an interface to handle `connection` and `message` events, assisting in the implementation of methods.  
- A `handleException` module was added on `@2060.io/service-agent-client` to the client library for exception handling in controllers. (**Consideration:** Is it better to use this here or in the `nestjs-client` library?)
- A chatbot model was created in the `examples/demo-dts` directory. This model includes basic modules for use when creating other projects (such as environment variable management via a configuration file, language support, basic event handling, and usage of the `nestjs-client` library).


packages/nestjs-client/src/messages/message.service.ts
```diff
@@ -0,0 +1,54 @@
...
+  ) {
+    this.url = options.url
+    this.version = options.version
+
+    this.apiClient = new ApiClient(this.url, this.version)
+
+    this.logger.debug(`Initialized with url: ${this.url}, version: ${this.version}`)
+  }
+
+  async received(event: MessageReceived): Promise<any> {
```
genaris, line 29: Why don't simply return a Promise<void>? Are you expecting to return anything else?

examples/demo-dts/Dockerfile
```diff
@@ -0,0 +1,46 @@
+FROM node:18 as base
```
genaris, line 1: Can we update base node images? Current [active LTS version](https://nodejs.org/en/about/previous-releases) is 22

packages/nestjs-client/src/messages/message.service.ts
```diff
@@ -0,0 +1,54 @@
...
+    await this.apiClient.messages.send(body)
+
+    if (this.eventHandler) {
+      await this.eventHandler.inputMessage(event.message)
+    }
+
+    return null
+  }
+
+  async updated(event: MessageStateUpdated): Promise<any> {
```
genaris, line 51: Same here?

PR #47, opened 2024-11-27
feat: Node.js Base Image Update, Demo DTS Refactor, and Documentation Improvements

lotharking:
Features:
- Service agent client: update logger
- Demo dts: 
  - update node image to `node:22-alpine`
  - add swagger
  - add used to `@nestjs/config'` for environments variables
  - create a core module and service for the main process
  - updated basic `inputMessage` method
  - create methods for handling the default chatbot
- Nestjs-client:
  - add swagger for basic implementations


examples/demo-dts/src/config/app.config.ts
```diff
@@ -11,40 +12,40 @@ export default registerAs('appConfig', () => ({
...
+  appPort: parseInt(process.env.AGENT_PORT) || 5000,
 
   /**
    * Hostname or IP address for the PostgreSQL database.
-   * Defaults to an empty string if POSTGRES_HOST is not set in the environment variables.
+   * Defaults 'postgres' string if POSTGRES_HOST is not set in the environment variables.
    * @type {string}
    */
-  postgresHost: process.env.POSTGRES_HOST || '',
+  postgresHost: process.env.POSTGRES_HOST || 'postgres',
```
genaris, line 22: I understand that for Unic.ID Issuer DTS we may want to force using Postgres backend, but do you think we need to force it also in the demo DTS?

In my opinion, it is desirable to keep the number of required components to the minimum. Of course, it is good to have these postgres-related variables because it is good to facilitate the creation of more robust DTS based on this example.

genaris (commented):
This PR involves changes in different topics:

- Updates base image to node-alpine
- Updates example demo DTS structure and fixes action menu handling
- Improves documentation for some controllers

While it is better to put them in different PRs, I'm not against on merging this one. But what does "fix: handle message select" refer to? Can we try to give it a more significant name?

examples/demo-dts/src/config/app.config.ts
lotharking: IMHO,  it would be better to keep the PostgreSQL configuration as it is currently recommended for our ecosystem's functionality. While it's true that making it configurable for future use is a good idea, I believe it would be worth leaving it as an issue to address later when the need arises to adapt to additional databases.

examples/demo-dts/src/core.module.ts
```diff
@@ -0,0 +1,35 @@
...
+    TypeOrmModule.forFeature([ConnectionEntity, SessionEntity]),
+    TypeOrmModule.forRootAsync({
+      imports: [ConfigModule],
+      useFactory: async (configService: ConfigService): Promise<TypeOrmModuleOptions> => ({
+        type: 'postgres',
+        host: configService.get<string>('appConfig.postgresHost'),
+        port: 5432,
+        username: configService.get<string>('appConfig.postgresUser'),
+        password: configService.get<string>('appConfig.postgresPassword'),
+        database: configService.get<string>('appConfig.postgresUser'),
```
genaris, line 20: What about calling this postgresDatabaseName?

examples/demo-dts/src/config/app.config.ts
genaris: I was actually referring to leaving it blank like we do in Service Agent to use embedded SQLite database instead of an external one. This way, we can set up a lightweight development demo composed by only two pieces: Service Agent and chatbot DTS backend (i.e. no need of postgres and redis).

In this case, I think this can be achieved by allowing the specification of a database type (`type` parameter in TypeORM), so a `databaseType` environment variable can be added. I guess this will require to add SQLite driver as a dependency for the demo.

But no problem on leaving it for another PR.

genaris (approved):

examples/demo-dts/src/core.module.ts
lotharking: fixed: added `POSTGRES_DB_NAME` variable and `postgresDbName` in the config of nestjs

PR #50, opened 2024-12-02
feat: adjust params on call offer message

lotharking:


examples/chatbot/docker-compose.yml
```diff
@@ -15,9 +15,12 @@ services:
       - AGENT_PUBLIC_DID=did:web:10.82.14.12%3A3000
       - AGENT_ENDPOINT=ws://10.82.14.12:3001
       - AGENT_INVITATION_IMAGE_URL=http://10.82.14.12:3001/avatar.png
-      - AGENT_NAME=Chatbot Agent
+      - AGENT_LABEL=Chatbot Agent
       - USE_CORS=true
       - EVENTS_BASE_URL=http://chatbot-backend:5000
+      - AGENT_WALLET_KEY_DERIVATION_METHOD=ARGON2I_INT
```
genaris, line 21: I think it is better to leave it blank so it takes the default one (ARGON2I_MOD)

examples/chatbot/index.ts
```diff
@@ -361,13 +361,10 @@ expressHandler.messageReceived(async (req, res) => {
       try {
         if (!wsUrl || !roomId) {
           // Create a room if not given as argument
-          const result = await fetch(`${WEBRTC_SERVER_BASE_URL}/rooms`, {
-            headers: {
-              Accept: 'application/json',
-              'Content-Type': 'application/json',
-            },
+          const result = await fetch(`${WEBRTC_SERVER_BASE_URL}/rooms/${randomUUID()}`, {
```
genaris, line 364: Why did you need to specify this parameter? As far as I know, it works without it, doesn't it?

examples/chatbot/index.ts
```diff
@@ -361,13 +361,10 @@ expressHandler.messageReceived(async (req, res) => {
...
             method: 'POST',
-            body: JSON.stringify({ maxPeerCount: 2, eventNotificationUri: `${PUBLIC_BASE_URL}/call-events` }),
+            headers: { 'Content-Type': 'application/json' },
```
genaris, line 366: It seems to be unrelated changes to this PR. Not sure why they were needed.

packages/model/src/messages/calls/CallOfferRequestMessage.ts
```diff
@@ -16,13 +21,31 @@ export class CallOfferRequestMessage extends BaseMessage {
...
+      this.offerStartTime = options.offerStartTime
+      this.description = options.description
       this.parameters = options.parameters
     }
   }
 
   public readonly type = CallOfferRequestMessage.type
   public static readonly type = MessageType.CallOfferRequestMessage
 
+  @Expose()
```
genaris, line 34: Note that these @Expose() decorators are not actually needed, since all fields will be exposed by default.

genaris (changes requested):

genaris (approved):

PR #51, opened 2024-12-02
feat: decline support in MRZ and eMRTD requests

genaris:
Add support for user refusal in MRZ and eMRTD data request flows. Timeout errors from server are defined but not yet supported, since we probably we won't need it initially on UnicID flows (as we simply send text messages and contextual menu items).


yarn.lock
```diff
@@ -23,6 +23,18 @@
     mrz "^4.2.0"
     tsyringe "^4.8.0"
 
+"@2060.io/credo-ts-didcomm-mrtd@^0.0.9":
```
lotharking, line 26: It looks good, the only thing missing is the update of MRTD v0.0.9 in the service agent models, so that the yarn lock doesn't manage two versions of the same library.

lotharking (approved):

PR #54, opened 2024-12-09
feat: Add ISO 3166-1 support for nationality and issuingState, and restrict sex to F, M, or X

lotharking:
Feature:
- Extracts sex from MRZ, defaults to 'X' if missing. Validates against allowed values ('male', 'female'), converts to 'M', 'F', or 'X'.
- Maps nationality and issuingState to ISO 3166-1 alpha-2 code if available, otherwise defaults to undefined.

### DIFF
--- a/packages/model/package.json
+++ b/packages/model/package.json
@@ -23,6 +23,7 @@
     "class-transformer": "0.5.1",
     "class-validator": "0.14.1",
     "credo-ts-receipts": "^0.0.1-alpha.5",
+    "iso-3166-1": "^2.1.1",
     "mrz": "^4.2.0"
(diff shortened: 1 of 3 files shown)


genaris (commented):
ISO 3166 alpha-2 is a requirement we currently have in UnicID Issuer DTS. Maybe this should be addressed there, and keep it here like it comes from the passport.

In regards to gender I'd say the same. However I don't know what it does currently set in case of 'X'. Is it 'unspecified' or something similar?

lotharking:
@genaris At this moment, the library returns 'nonspecified' when is not recognize

Issue #60, opened 2024-12-11
Support ref and callbackUrl in presentation-request endpoint


genaris:
In order to make it possible to create backend for the Generic Verifier DTS for [UnicID], Service Agent must add support for asynchronous callbacks in the Presentation Request flows.

This will mainly consist on adding `ref` and `callbackUrl` to `invitation/presentation-request` endpoint so Service Agent client can keep track of the results of the presentation request in a simple manner (i.e. without the need to conform to the whole Service Agent API, which could be cumbersome for this simple purpose).

PR #61, opened 2024-12-12
feat: add presentation callback parameters

genaris:
## Summary

Presentation Request endpoint now adds two fields specified in [Unic.ID Generic verifier](https://github.com/2060-io/unic.id-dts-specs/blob/main/spec.md#generic-verifier-backend-generic-verifier-dts---simple-verifier-backend): `ref` and `callbackUrl`. 

These values are stored in the flow metadata, so once the flow is finished, callback URL is called.

## Changes
- 

## Related Issues
Solves #60 

## Checklist


doc/service-agent-api.md
```diff
@@ -772,13 +783,17 @@ It's a POST to `/invitation/presentation-request` which receives a JSON object i
 }
 ```
 
+`callbackUrl` is an URL that will be called by Service Agent when the flow completes. The request follows the [Presentation Callback API](#presentation-callback-api).
+
+`ref` is an optional, arbitrary string that will be included in the body of the request to the callback URL.
```
lotharking, line 788: On [[this spect]](https://github.com/2060-io/unic.id-dts-specs/blob/main/spec.md#generic-verifier-backend-api-1-1-generate-invitation-parameters), the `ref` field must be marked as a `mandatory` field.

examples/verifier/Dockerfile
```diff
@@ -0,0 +1,39 @@
+FROM node:18 as base
```
lotharking, line 1: `node:22-alpine` also works fine

lotharking (approved):

Issue #58, opened 2024-12-11
Support revocable credentials


genaris:
Support issuance of revocable AnonCreds credentials. This requires to set-up a Tails server and add the appropriate endpoints to support revocation registries and revocation status lists.

Service Agent will need also to create and maintain revocation registries, setting current revocation index appropriately. When current revocation registry is full, it should automatically create a new one.

Ideally, this should be done in an unattended way, using some default parameters (such as the number of records for each revocation registry) that could be overriden.

genaris:
Moved back to 'In progress' until #64 is finished.

PR #64, opened 2024-12-16
feat: Refactor AppModule and Add Credential Management with Revocation Support

lotharking:
#### Flow:
1. **Create a Credential**  
   - The user can request to create a credential.  
   - The library validates whether the credential can be created and ensures there isn't already one associated with the same hash.  
     - If a credential with the same hash exists, the credential will be revoked and associated with the new connection (from where the request is sent). A new credential request will then be initiated, ensuring the proper threadId is sent as part of the process
     - If valid, the credential creation request is sent.

2. **Rejection**  
   - Associates the thread and marks the credential as revoked (revoked only in terms of database nomenclature).  

3. **Acceptance**  
   - Associates the thread with the credential.

4. **Revocation**  
   - When a credential is revoked, it is marked as revoked in the database.

#### Additional Constraints:

- In all cases, a new record is created, ensuring indices do not overlap and atomicity is guaranteed during queries.  
- Records are expected to be assigned from `0` to the maximum value minus 1 (`-1`).  
  - Validation has confirmed that assignments from `1` to the maximum value fail.  
- Two revocation records will always remain available.  

## Changes
- A refactoring was implemented in the `AppModule` of the `nestjs-client` module. This was necessary because the original main module was not optimized for the recurrent use of variables, requiring multiple instantiations of the same variables. Modifying the `AppModule` and the configuration approach avoided redundancy when the module is instantiated. Compatibility with individual configurations was maintained.
- Support for credential management and revocation was implemented when configuring the `nestjs-client` module. The methods are controlled through the library, allowing for handling rejections and revocations as well.
- A demo was created to showcase how to use the `demo-dts`. As observed, the configuration required for complete credential management is minimal.

## Related Issues
- https://github.com/2060-io/2060-service-agent/issues/58

## Checklist


packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+
+  /**
+   * Ensures the creation and initialization of a credential definition with support for revocation registries.
+   *
+   * When this method is called:
+   * - Two default lists of possible revocation registries will be created. (only with supportRevocation)
+   * - This guarantees seamless handling in case one of the registries runs out of revocation capacity. (only with supportRevocation)
+   *
+   * **Usage Recommendation:**
+   * It is recommended to call this method inside an `OnModuleInit` lifecycle hook to validate the existence
```
genaris, line 50: This is only for the specific case where a service is intended to issue only a single type of credential. It can be useful in those cases but in my opinion is too particular to name it as a general practice.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * @param name - The name of the credential definition. Defaults to "Chatbot" if not provided.
+   * @param version - The version of the credential definition. Defaults to "1.0" if not provided.
+   * @param attributes - A list of attributes that the credential definition supports.
+   * @param supportRevocation - Whether revocation is supported. Defaults to `false` if not provided.
+   * @param maximumCredentialNumber - The maximum number of credentials. Defaults to `1000` if not provided.
+   *
+   * @returns A promise that resolves when the credential definition and revocation registries are created.
+   */
+  async create(
```
genaris, line 61: This method creates a credential definition (or "credential type", as we call it in Service Agent). We need to do a clear distinction between a credential type and the credentials issued using it as a base.

It might be more appropriate to use another module/service for that, but if we keep it in a single module for the sake of simplicity, we can maybe call this method `createType` or something related.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+      name?: string
+      version?: string
+      supportRevocation?: boolean
+      maximumCredentialNumber?: number
+      autoRevocationEnabled?: boolean
+    } = {},
+  ) {
+    const { name, version, supportRevocation, maximumCredentialNumber, autoRevocationEnabled } = options
+
+    const [credential] = await this.apiClient.credentialTypes.getAll()
```
genaris, line 73: Again: this assumes that every service is supposed to define a single credential type. This is not correct.

If you want to know if a given credential type is already created, you should query credential types by name and version.

genaris (changes requested):
Looking good. Some comments regarding the 'reference identifier' and supporting having more than a single credential type.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+   *   { name: "email", value: "john.doe@example.com" },
+   *   { name: "name", value: "John Doe" },
+   * }
+   *
+   * @returns {Promise<void>} A promise that resolves when the credential issuance is successfully
+   * sent. If an error occurs during the process, the promise will be rejected.
+   */
+  async issuance(
+    connectionId: string,
+    records: Record<string, any>,
```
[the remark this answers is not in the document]
lotharking: fix: added claims on issue method

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+    options?: { identifier?: string },
+  ): Promise<void> {
+    const hashIdentifier = options?.identifier ? this.hashIdentifier(options.identifier) : null
+    const credentials = await this.apiClient.credentialTypes.getAll()
```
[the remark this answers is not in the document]
lotharking: fix: add optional credentialDefinitionId on issue method

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: implement selection credential by name and version

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,319 @@
...
+   *
+   * @returns {Promise<void>} A promise that resolves when the credential issuance is successfully
+   * sent. If an error occurs during the process, the promise will be rejected.
+   */
+  async issuance(
+    connectionId: string,
+    records: Record<string, any>,
+    options?: { identifier?: string },
+  ): Promise<void> {
+    const hashIdentifier = options?.identifier ? this.hashIdentifier(options.identifier) : null
```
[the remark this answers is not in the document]
lotharking: docs: update

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: updated name method create

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: updated for many credentials

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,331 @@
...
+   * @param {object} [options] - Additional options for credential issuance.
+   *
+   * ### Options
+   * - `refId` (optional, `string`): A unique identifier for the credential. If provided:
+   *   - Ensures the credential is unique and any existing credential with the same `refId` is revoked.
+   *   - Used for managing unique credentials like official documents.
+   *   - Encrypted in the database for security.
+   * - `credentialDefinitionId` (optional, `string`): Specifies the ID of the credential definition to use.
+   *   - If not provided, the first available credential definition is used.
+   * - `autoRevocationEnabled` (optional, `boolean`): Whether automatic revocation is enabled (default false)
```
genaris, line 111: I think this should be more self-explanatory. Maybe something like: `revokeIfAlreadyIssued`

packages/nestjs-client/src/credentials/credential.entity.ts
```diff
@@ -0,0 +1,34 @@
...
+  revocationRegistryIndex?: number
+
+  @Column({ type: 'varchar', nullable: true })
+  connectionId?: string
+
+  @Column({ type: 'varchar', nullable: true })
+  threadId?: string
+
+  @Column({ type: 'varchar', nullable: true })
+  hashIdentifier?: string
```
genaris, line 24: I think this field should be renamed. `refIdHash`?

genaris (changes requested):

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,331 @@
...
+      const credential = await this.apiClient.credentialTypes.create({
+        id: utils.uuid(),
+        name,
+        version,
+        attributes,
+        supportRevocation,
+      })
+
+      await this.saveCredential(credential.id, supportRevocation)
+      await this.saveCredential(credential.id, supportRevocation)
```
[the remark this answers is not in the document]
lotharking: fix: name credentialType

packages/nestjs-client/src/credentials/credential.entity.ts
lotharking: fix: name refIdHash

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: add revokeIfAlreadyIssued

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,331 @@
...
+   * const claims = [
+   *   { name: "email", value: "john.doe@example.com" },
+   *   { name: "name", value: "John Doe" }
+   * ];
+   * ```
+   *
+   * @param {object} [options] - Additional options for credential issuance.
+   *
+   * ### Options
+   * - `refId` (optional, `string`): A unique identifier for the credential. If provided:
```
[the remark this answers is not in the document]
lotharking: docs: update

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+  }
+
+  /**
+   * Accepts a credential by associating it with the provided thread ID.
+   * @param connectionId - The connection ID associated with the credential.
+   * @param threadId - The thread ID to link with the credential.
+   * @throws Error if no credential is found with the specified connection ID.
+   */
+  async accept(connectionId: string, threadId: string): Promise<void> {
+    const cred = await this.credentialRepository.findOne({
```
genaris, line 244: A problem I see here is that you are using `connectionId` to identify the credential the event is related to. This is not correct, since you can issue multiple credentials to a single DIDComm connection.

The proper way you should identify a credential is by its threadId. Therefore, these `accept` and `reject` methods should only receive the threadId.

You can find the threadId in the response of the Message endpoint (the `id` correspond to the threadId).

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+      where: {
+        connectionId,
+        revoked: false,
+      },
+      order: { createdTs: 'DESC' },
+    })
+    if (!cred) throw new Error(`Credencial with connectionId ${connectionId} not found.`)
+
+    cred.threadId = threadId
+    cred.revoked = true
```
genaris, line 274: A rejection by the user does not mean that a credential has been revoked. Maybe you can create a `status` field for the credential entity that covers the entire life cycle of a credential: probably offered, accepted, rejected and revoked may be enough.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+  async createType(
+    attributes: string[],
+    options: {
+      name?: string
+      version?: string
+      supportRevocation?: boolean
+      maximumCredentialNumber?: number
+    } = {},
+  ) {
+    const { name = 'Chatbot', version = '1.0', supportRevocation, maximumCredentialNumber } = options
```
genaris, line 65: I don't think it is correct to leave name and version as optional fields.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+            connectionId,
+            credentialDefinitionId,
+            ...(refIdHash ? { refIdHash } : {}),
+          })
+          return {
+            revocationRegistryDefinitionId: undefined,
+            revocationRegistryIndex: undefined,
+          }
+        }
+        const invalidRegistries = await transaction.find(CredentialEntity, {
```
genaris, line 167: This logic is quite obscure to me. I don't think it will work.

I guess that somewhere (maybe another table/entity) you should keep track of the current revocation registry and index for each credential definition, and use that as the source of truth to assign the revocation registry/index of each issued credential, and tell if it is needed to create a new revocation registry or not.

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+   */
+  async reject(connectionId: string, threadId: string): Promise<void> {
+    const cred = await this.credentialRepository.findOne({
+      where: {
+        connectionId,
+        revoked: false,
+      },
+      order: { createdTs: 'DESC' },
+    })
+    if (!cred) throw new Error(`Credencial with connectionId ${connectionId} not found.`)
```
genaris, line 271:
```suggestion
    if (!cred) throw new Error(`Credential with connectionId ${connectionId} not found.`)
```

As explained before, threadId should be used instead of connectionId.

The connectionId is only needed to determine where to send the notification in case of a revocation.

genaris (changes requested):

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+      )
+    await this.credentialRepository.save({
+      credentialDefinitionId,
+      revocationDefinitionId: revocationRegistry,
+      revocationRegistryIndex: this.maximumCredentialNumber,
+    })
+    return revocationRegistry
+  }
+
+  private refIdHash(refId: string): string {
```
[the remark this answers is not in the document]
lotharking: fix: change name refIdHash by hash

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: remove name and version from optionals fields

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    const credentialType =
+      credentialTypes.find(credType => credType.id === options?.credentialDefinitionId) ?? credentialTypes[0]
+    if (!credentialType) {
+      throw new Error(
+        'No credential definitions found. Please configure a credential using the create method before proceeding.',
+      )
+    }
+    const { id: credentialDefinitionId, revocationSupported } = credentialType
+
+    const cred = await this.credentialRepository.findOne({
```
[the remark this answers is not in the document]
lotharking: fix: implement revocation for all non-revoked credentials

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+        new CredentialRevocationMessage({
+          connectionId: cred.connectionId,
+          threadId: cred?.threadId,
+        }),
+      ))
+    this.logger.log(`Revoke Credential: ${cred.id}`)
+  }
+
+  // private methods
+  private async saveCredentialType(
```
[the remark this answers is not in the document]
lotharking: fix: updated to createRevocationRegistry

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,337 @@
...
+    this.logger.debug('sendCredential with claims: ' + JSON.stringify(claims))
+  }
+
+  /**
+   * Accepts a credential by associating it with the provided thread ID.
+   * @param connectionId - The connection ID associated with the credential.
+   * @param threadId - The thread ID to link with the credential.
+   * @throws Error if no credential is found with the specified connection ID.
+   */
+  async accept(connectionId: string, threadId: string): Promise<void> {
```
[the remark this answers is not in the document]
lotharking: fix: change name to handleAcceptance/handleRejection

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: acept and reject only by thread

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: fix: implement to CredentialStatus

packages/nestjs-client/src/credentials/credential.service.ts
lotharking: feat: Creation of the RevocationRegistryEntity for managing revocation records.

Please review this last change with special care. The `RevocationRegistryEntity` was created, and during its implementation, it was detected that the `threadId` returned by the `CredentialIssuanceMessage` was not being properly handled by the SA. As a solution, `GenericRecords` was used to retrieve it. Additionally, as you mentioned, the handling of credentials has been improved by dividing the entities.

packages/nestjs-client/src/credentials/revocation-registry.entity.ts
```diff
@@ -0,0 +1,22 @@
+import { Entity, Column, PrimaryGeneratedColumn, CreateDateColumn, UpdateDateColumn } from 'typeorm'
+
+@Entity('revocation_registries')
+export class RevocationRegistryEntity {
```
genaris, line 4: Add credentialDefinitionId

packages/nestjs-client/src/credentials/revocation-registry.entity.ts
```diff
@@ -0,0 +1,22 @@
...
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @Column({ type: 'varchar', nullable: false })
+  revocationDefinitionId!: string
+
+  @Column({ type: 'integer', nullable: false, default: 0 })
+  revocationRegistryIndex!: number
```
genaris, line 12:
```suggestion
  currentIndex!: number
```

packages/nestjs-client/src/credentials/credential.entity.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { RevocationRegistryEntity } from './revocation-registry.entity'
+
+@Entity('credentials')
+export class CredentialEntity {
+  @PrimaryGeneratedColumn('uuid')
+  id!: string
+
+  @Column({ type: 'varchar', nullable: false })
+  credentialDefinitionId!: string
```
genaris, line 21: add revocationRegistryIndex

packages/nestjs-client/src/credentials/credential.service.ts
```diff
@@ -0,0 +1,311 @@
...
+    const cred: CredentialEntity = await this.entityManager.transaction(async transaction => {
+      if (!revocationSupported) {
+        return await transaction.save(CredentialEntity, {
+          connectionId,
+          credentialDefinitionId,
+          ...(refIdHash ? { refIdHash } : {}),
+        })
+      }
+      let lastCred = await transaction
+        .createQueryBuilder(RevocationRegistryEntity, 'registry')
```
genaris, line 164: add credentialDefinitionId to the filter

genaris (changes requested):

packages/nestjs-client/src/credentials/credential.entity.ts
lotharking: fix: add credentialDefinitionId to revocation registry entity

packages/nestjs-client/src/credentials/revocation-registry.entity.ts
lotharking: fix: implemented currentIndex

packages/nestjs-client/src/credentials/revocation-registry.entity.ts
lotharking: fixed

genaris (approved):

PR #71, opened 2025-01-22
fix: add mrzData field to EMrtdDataSubmitMessage

lotharking:
## Summary
As requested in [issue #40](https://github.com/2060-io/unic.id-issuer-dts/issues/40), the `mrzData` field has been added to the response of the `EMrtdDataSubmitMessage` to use it as a unique hash. 

This enhancement leverages new features in the Service Agent to support the issuance of revocable credentials. According to the specification, the Issuer DTS must store all credential issuances based on a hash of the MRZ used to generate them. 

## Changes
- 

## Checklist
- [x] I have flagged specific areas for further review, if necessary.


packages/model/src/messages/mrtd/EMrtdDataSubmitMessage.ts
```diff
@@ -31,6 +31,7 @@ export type EMrtdRawData = {
     title?: string
     personalSummary?: string
     custodyInformation?: string
+    mrzData?: string
```
genaris, line 34: Since we do already have a field named `mrzOptionaldata`, I think it would be better if we call this field `mrzString` to make it clear that it is the full, non-parsed, string

genaris (commented):
I think it is not a fix but rather a feat.

packages/model/src/messages/mrtd/EMrtdDataSubmitMessage.ts
lotharking: updated name

genaris (approved):

PR #72, opened 2025-01-22
feat: add stats module

lotharking:
## Summary
The **StatsModule** is created to enable the use of statistics. Additionally, the **StatProducerService** is implemented, which provides the `spool` method to add new statistics.
## Changes
- 

## Checklist


packages/nestjs-client/src/jms/stats-producer.service.ts
```diff
@@ -0,0 +1,155 @@
...
+    username: string
+    password: string
+    reconnectLimit: number
+    threads: number
+    delay: number
+  }
+
+  // The constructor initializes the service with default settings suitable for
+  // a local development environment. By default, it points to the localhost of
+  // the 2060 project (`https://github.com/mobiera/stats`).
```
genaris, line 28: Is this link OK? You mention "2060 project" and then put a link to mobiera repo

packages/nestjs-client/README.md
```diff
@@ -70,6 +71,21 @@ The `nestjs-client` allows dynamic configuration through various module options
 - `imports`: An array of additional modules to import, such as service modules or other shared functionality.
 - `url`: The URL of the service agent for connecting to it.
 - `version`: Specifies the version of the service agent API to use.
+- `jmsOptions`: Configuration options for the **JMS (Java Message Service) broker**, including host, port, queue name, authentication, and reconnection settings. The default broker used is **Apache Artemis**.
```
genaris, line 74: I think JMS is nowadays called "Jakarta Message Service". Unless we are using an older version here. 

BTW, here you directly add these `jmsOptions` into root configuration. Could it (in practice) be used for anything else than stats? If not, I think it could go under an statsConfig object or something like that.

It is important to make it clear what this JMS is used for. Otherwise it will be confusing for users that might not need to use any stats at all.

genaris (approved):
I'm not a big fan of using this older JMS implementation, but if it can save us some time it it OK for me!

lotharking:
> I'm not a big fan of using this older JMS implementation, but if it can save us some time it it OK for me!

While it is true that JMS is an old and stable implementation that is only natively supported in Java, Java services are still its primary means of communication for message delivery. At the time, implementing/using a statistics module was a great idea (nowadays, there are several open-source tools available). Grafana is a widely used option.

packages/nestjs-client/README.md
lotharking: This configuration is useful not only for the main statistics module currently in use. While exploring alternative solutions could be a great idea, **JMS is still present in older chatbots** like Friend Chat and other legacy systems.  

When migrating these systems, **JMS can help smooth the testing process**, ensuring a more seamless transition. For example, when migrating the necessary components from **2060 to Node.js**, **JMS could be gradually deprecated** instead of being removed abruptly.

packages/nestjs-client/src/jms/stats-producer.service.ts
lotharking: fixed

PR #78, opened 2025-02-07
feat: add waiting state

lotharking:
## Summary
**Related issue:** https://github.com/2060-io/unic.id-generic-verifier-dts/issues/9

## Changes
- 

## Checklist

### DIFF
--- a/packages/main/src/events/CallbackEvent.ts
+++ b/packages/main/src/events/CallbackEvent.ts
@@ -5,6 +5,7 @@ import { TsLogger } from '../utils/logger'
 
 export enum PresentationStatus {
   OK = 'ok',
+  WAITING = 'waiting',
   REFUSED = 'refused',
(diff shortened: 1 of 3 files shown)


packages/main/src/events/CallbackEvent.ts
```diff
@@ -5,6 +5,7 @@ import { TsLogger } from '../utils/logger'
 
 export enum PresentationStatus {
   OK = 'ok',
+  WAITING = 'waiting',
```
genaris, line 8:
```suggestion
  CONNECTED = 'connected',
```

I think probably "Connected" is a bit more meaningful here and gives us the opportunity of adding other states we may need in the future (e.g. "CODE_SCANNED" or something)

genaris (approved):
This is a very elegant and simple solution! Let's see how this new feature improves the UX of the verifier.

PR #91, opened 2025-03-10
fix: implement cors

lotharking:
## Changes
- 

## Checklist


genaris (approved):
I'm OK with restricting CORS depending on server config but what is the reason behind this update? Was there any issue with previous parameters?

lotharking:
The recently detected CORS issue in the verifier is due to improper CORS handling in the agent. Since the default configuration does not allow it in these cases, it is necessary to explicitly [enable CORS according to the configuration.
](https://thriveread.com/how-to-enable-nestjs-cors/)

PR #92, opened 2025-03-25
fix: invitation to public services using pthid

genaris:
BREAKING CHANGE: this generates an invitation that is not compatible with Hologram App <= 2.6.1. 

It will be good to merge it only when Hologram 2.6.2/2.7.0 has been published.

## Changes
- 

## Checklist


lotharking:
Maybe the appropriate name could be `fix!` instead of `fix:` so that the release process can detect the breaking change.

genaris:
> Maybe the appropriate name could be fix! instead of fix: so that the release process can detect the breaking change.

Yes, but I'm afraid that current release config will imply this breaking change to update the verison to 2.0.0 and I don't think it's worth for such a small change like this one.

lotharking (approved):

PR #93, opened 2025-04-28
fix: update credential phone

lotharking:
## Changes
- 

## Checklist


genaris (commented):
@lotharking what if we add the JSON as an environment variable or override the file in each deployment (e.g. by mounting it as a volume in certain path)? I think we did it in other projects like the didcomm mediator for firebase cfg, didn't we @gabrielmatau79 ?

The challenge here would be to create an script or something that will make easy for everyone to generate such file without the need of deploying the whole thing.

lotharking:
@genaris the issue has been [resolved](https://github.com/2060-io/2060-demos-deploy-dev/pull/16) as we discussed. The file replacement was implemented and it's now working correctly. Thanks for your support.

PR #104, opened 2025-05-20
fix: use the short url to connections notify

lotharking:
## Changes
- 

## Checklist


packages/main/src/invitationRoutes.ts
```diff
@@ -25,6 +27,25 @@ export const addInvitationRoutes = async (app: express.Express, agent: ServiceAg
         return
       }
 
+      const connRecord = shortUrlRecord?.getTag('relatedFlowId') as string
```
genaris, line 30: Move to (req.accepts('json')) { } in order to be only catched when Hologram App actually scans the QR code (otherwise, it might be possible that a web browser does that, e.g. in [http:/](https://hologram.zone/?_url=aHR0cHM6Ly9hLnVuaWMtaWQtdmVyaWZpZXIuZGVtb3MuZGV2LjIwNjAuaW8vcz9pZD0wMzVjNjVjNC0yZDM5LTRlMDQtOGFkMC05MDUwMDFhZmFmMjU))

packages/main/src/events/ConnectionEvents.ts
```diff
@@ -47,33 +46,6 @@ export const connectionEvents = async (agent: ServiceAgent, config: ServerConfig
           })
       }
```
genaris, line 48: Probably it worths to keep this CONNECTED status and add another one for SCANNED

lotharking: Hey @genaris, the new state has been created on https://github.com/2060-io/generic-verifier/pull/37

genaris (approved):

PR #141, opened 2025-06-26
chore: update repo structure

genaris:
Changing to the more typical monorepo structure, with:

- apps (vs-agent)
- packages  (clients, model)


apps/vs-agent/Dockerfile
```diff
@@ -10,7 +10,7 @@ ENV RUN_MODE="docker"
 COPY package.json pnpm-lock.yaml pnpm-workspace.yaml .npmrc ./
 
 COPY packages/model/package.json packages/model/package.json
-COPY packages/main/package.json packages/main/package.json
+COPY apps/vs-agent/package.json apps/vs-agent/package.json
```
lotharking, line 13: We used to use `packages/main` because the Dockerfile was at the root of the monorepo. If it has now been moved to the `apps/vs-agent` directory, the Docker Compose file must be updated to include the `dockerfile` parameter parameter to point to the new directory.

lotharking (approved):

Issue #151, opened 2025-07-03
Remove/deprecate redundant config parameters


genaris:
VS Agent has a massive amount of environment variables, and some of them aren't actually necessary or can be solved in another way. Some comments:

- AGENT_ENDPOINT: we can deprecate it in favour of AGENT_ENDPOINTS. If somebody uses this older setting, we simply reconstruct the array with a single element (pretty much the same as it works right now, but we need to explicitly deprecate it in the documentation and VS-A logs)
- ENABLE_WS and ENABLE_HTTP: we can remove them and enable/disable them depending on the items in AGENT_ENDPOINTS
- AGENT_NAME: it is deprecated since a long time now. We can remove it and show an error if set, asking to replace it by AGENT_WALLET_ID and AGENT_WALLET_KEY
- ANONCREDS_SERVICE_BASE_URL: we can remove it. It is always the same as PUBLIC_API_BASE_URL (AnonCreds endpoitns will be always exposed, but I don't see an issue about that)

lotharking:
AGENT_WALLET_KEY_DERIVATION_METHOD: has been deprecated

Issue #154, opened 2025-07-03
Add README for examples


genaris:
NestJS Client Demo does include one, which is probably too long. The other ones don't have anything explaining how to set them up and what do they do.

PR #157, opened 2025-07-03
docs: organize env vars in vs agent README

genaris:
In this PR I'm assuming that we are going to deprecate/remove some variables, as stated in #151

### DIFF
--- a/apps/vs-agent/src/invitationRoutes.ts
+++ b/apps/vs-agent/src/invitationRoutes.ts
@@ -8,7 +8,7 @@ import { VsAgent } from './utils/VsAgent'
 import { createInvitation } from './utils/agent'
 import { TsLogger } from './utils/logger'
 
-// Add invitation endpoints (TODO: remove as it should be part of an external API)
+// Add invitation endpoints
 export const addInvitationRoutes = async (app: express.Express, agent: VsAgent) => {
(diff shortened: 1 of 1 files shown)


apps/vs-agent/README.md
```diff
@@ -1,37 +1,54 @@
...
+# VS Agent
 
-2060 Agent running in a container, used as gateway for any service to access 2060 ecosystem.
+Verifiable Service Agent is a web application that allows to create Verifiable Services.
 
 ## Configuration
 
-At the moment, all configuration is done by environment variables. All of them are optional for development
-but likely needed for production and test deployments.
+Most configuration of VS all configuration is done by environment variables. All of them are optional for development
```
lotharking, line 7: I'm not sure about this. I think a good practice would be to specify which environment variables are required for a basic, common configuration, and include a small example with the recommended values (To enable faster testing for users).

genaris: This is just an introductory paragraph (that is not well written, by the way). If you see the subsections, there is an explanation about variables for a basic configuration.

We'll have examples as part of #154 and Hologram docs website. I think we need to go step by step until we get the good balance of information between this repo and the one on the docs.

lotharking (approved):

PR #158, opened 2025-07-04
feat: default redirect in invitation endpoint to true

### LINKED ISSUE #156: set REDIRECT_DEFAULT_URL_TO_INVITATION_URL to true by default
Currently, if it's not defined it is treated as false.

genaris:
A small change in behaviour but useful for most deployments: `invitation` endpoint will redirect to `https://hologram.zone?=oob=xxx` by default when `REDIRECT_DEFAULT_URL_TO_INVITATION_URL` is not defined.

Closes #156

### DIFF
--- a/apps/vs-agent/src/invitationRoutes.ts
+++ b/apps/vs-agent/src/invitationRoutes.ts
@@ -8,7 +8,7 @@ import { VsAgent } from './utils/VsAgent'
 import { createInvitation } from './utils/agent'
 import { TsLogger } from './utils/logger'
 
-// Add invitation endpoints (TODO: remove as it should be part of an external API)
+// Add invitation endpoints
 export const addInvitationRoutes = async (app: express.Express, agent: VsAgent) => {
   // Retrieve the URL that corresponds to a given short URL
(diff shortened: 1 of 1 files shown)


apps/vs-agent/src/invitationRoutes.ts
```diff
@@ -59,7 +59,7 @@ export const addInvitationRoutes = async (app: express.Express, agent: VsAgent)
   // Generate a regular invitation
   app.get('/invitation', async (req, res) => {
     const { url: invitationUrl } = await createInvitation(agent)
-    if (process.env.REDIRECT_DEFAULT_URL_TO_INVITATION_URL === 'true') res.redirect(invitationUrl)
+    if (process.env.REDIRECT_DEFAULT_URL_TO_INVITATION_URL !== 'false') res.redirect(invitationUrl)
```
lotharking, line 62: The docs or another PR mention that the default is false, but with this change, it becomes true by default. Is that correct? #157

genaris: Changes in doc will be at #157

PR #167, opened 2025-07-08
feat: implement self vtr post method on admin port

lotharking:
## Changes
- 

## Checklist


apps/vs-agent/src/controllers/selft-vtr/selfVtrController.ts
genaris, line 1: There is a typo in the directory name

PR #169, opened 2025-07-09
feat: make public api server a nestjs app

lotharking:
## Changes
- 

## Checklist


apps/vs-agent/src/controllers/didcomm/did/DidWebRoutesController.ts
```diff
@@ -0,0 +1,262 @@
...
+}
+
+const fileStorage = diskStorage({
+  filename: (req: any, file: { originalname: string }, cb: (arg0: null, arg1: string) => void) => {
+    cb(null, file.originalname + '-' + new Date().toISOString())
+  },
+})
+
+@Controller()
+export class DidWebRoutesController {
```
genaris, line 61:
```suggestion
export class DidWebController {
```

apps/vs-agent/src/didcomm.module.ts
genaris, line 1: I think calling this 'didcomm' is inaccurate, since this is actually for the whole public API, including DIDComm. 'public' would be probably more appropiate.

apps/vs-agent/src/app.module.ts
genaris, line 1: This became now the 'admin' module, doesn't it?

apps/vs-agent/src/utils/setupAgent.ts
```diff
@@ -78,14 +80,11 @@ export const setupAgent = async ({
     publicApiBaseUrl,
   })
 
-  const app = express()
-
-  if (useCors) app.use(cors())
-
+  const app = await NestFactory.create(DidCommModule.register(agent))
```
genaris, line 83: I don't like this app being created in a different place than the admin one. I think it will be clearer if both are in `main.ts`. Do you think it is too complicated to do so? We can leave to another PR if you need help in setting up Credo's transports.

genaris (changes requested):

apps/vs-agent/src/app.module.ts
lotharking: yes, update the filename

apps/vs-agent/src/utils/setupAgent.ts
lotharking: Created https://github.com/2060-io/vs-agent/issues/195 for standar api creation

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -7,10 +7,10 @@ import { AnonCredsRequestedAttribute } from '@credo-ts/anoncreds'
 import { Controller, Get, Post, Body } from '@nestjs/common'
 import { ApiBody, ApiTags } from '@nestjs/swagger'
 
-import { PUBLIC_API_BASE_URL } from '../../config/constants'
-import { UrlShorteningService } from '../../services/UrlShorteningService'
-import { VsAgentService } from '../../services/VsAgentService'
-import { createInvitation } from '../../utils/agent'
+import { PUBLIC_API_BASE_URL } from '../../../config/constants'
```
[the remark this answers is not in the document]
lotharking: Added to ServerConfig since it now always has a value.

PR #175, opened 2025-07-15
feat: set public API and endpoints from public DID

genaris:
This simplifies the setup of 99% of Verifiable Services by creating DIDComm endpoints and public API base URL from `AGENT_PUBLIC_DID`, meaning that you won't need to explicitlly set `AGENT_ENDPOINTS` and `PUBLIC_API_BASE_URL` anymore.

It is still possible to define the older variables for two cases:

1) When we are in local development mode and we don't want to set a public DID (e.g. we are using http)
2) When we want to override defaults (e.g. we want to expose both ws and http DIDComm endpoints or even include another domain that redirects to ours)

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -4,22 +4,24 @@
   "separate-pull-requests": false,
   "bump-minor-pre-major": true,
   "bump-patch-for-minor-pre-major": true,
+  "release-type": "node",
   "packages": {
-    "apps/vs-agent": {
-      "release-type": "node"
-    },
-    "packages/model": {
-      "release-type": "node"
-    },
-    "packages/client": {
-      "release-type": "node"
-    },
-    "packages/nestjs-client": {
-      "release-type": "node"
-    },
-    "examples/chatbot": {
-      "release-type": "node"
-    }
+    "apps/vs-agent": { "component": "vs-agent" },
+    "packages/model": { "component": "model" },
+    "packages/client": { "component": "client" },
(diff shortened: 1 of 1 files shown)


apps/vs-agent/README.md
```diff
@@ -28,18 +26,17 @@ These variables are usually important for every deployment, since they define ho
...
-
-> **Note**: While not mandatory, it is recommended to set an agent public DID matching external hostname (e.g. if your VS-A instance pblic API is accessible in `https://myagent.com:3000` you must set AGENT_PUBLIC_DID to `did:web:myagent.com%3A3000`), which will make possible for the agent to create its own credential types and therefore issue credentials. Note that you'll need HTTPS in order to fully support did:web specification.
+> **Note**: Although it is possible to run VS Agent without any public DID, it is mandatory to do so in order to make possible for the agent to create its own credential types and therefore issue credentials. Note that you'll need HTTPS in order to fully support did:web specification.
 >
 > Public DID will be used also for agents to easily connect to it using DIDComm without the need of creating an explicit invitation by doing a GET request to `/invitation` endpoint.
+>
+> If you don't specify a public DID, you might set up `AGENT_PUBLIC_API_BASE_URL` and `AGENT_ENDPOINTS` manually.
```
lotharking, line 35:
```suggestion
> If you don't specify a public DID, you might set up `PUBLIC_API_BASE_URL` and `AGENT_ENDPOINTS` manually.
```

lotharking (approved):

PR #191, opened 2025-07-17
feat: add configurable parameters support for linked vp

lotharking:
**Summary:**

Hey @genaris,
the `SELF_VTR_ENABLED` variable has been replaced with the proposed `TESTVTR_ENABLE` for consistency in naming. I'm unsure whether we should also rename all related references (such as class names, constants, or filenames). Do you think we should fully migrate everything to the new name, or maintain partial compatibility with the existing ones?

Also, I added a helper function `urlToBase64` to handle image URL conversion. Would it make sense to always apply this by default to the `AGENT_INVITATION_IMAGE_URL`, or should we keep it limited to cases like these claim resolutions?

---

**Changes:**

* Removed `SELF_VTR_ENABLED` environment variable.
* Added new environment variables to make credential subject parameters configurable (e.g. service/org metadata).
* Removed the `POST /self-vtr/upload/:schemaId` endpoint since claim configuration is now handled through environment variables.
* Migrated previous validation logic from the `POST` endpoint to the internal default claim resolution process (inside `getClaims`).
* Created `urlToBase64()` function to fetch and convert remote images to base64 format.
* Applied `urlToBase64()` to the default claim logo, using the `AGENT_INVITATION_IMAGE_URL`.

Let me know your thoughts on the open questions and direction.

### DIFF
--- a/apps/vs-agent/README.md
+++ b/apps/vs-agent/README.md
@@ -111,7 +111,27 @@ VS-A fetches capabilities from the `discovery.json` file (which is located at at
 
 ### Self VR
 
-VS-A enable features related to the Verifiable Trust Registry ([Self VTR Configuration Guide](../../doc/self-tr-routes.md)).
+To enable the Self-Verifiable Trust Registry API endpoints, you must set the following environment variables in your `.env` file or system environment. These variables control the agent's identity, endpoints, and the data used for example credentials:
+
+| Variable                        | Description                                                      | Example Value                                  |
+|----------------------------------|------------------------------------------------------------------|------------------------------------------------|
+| `SELF_ISSUED_VTC_ORG_TYPE`               | Organization type for example credential                         | `PRIVATE`                                      |
+| `SELF_ISSUED_VTC_ORG_COUNTRYCODE`        | Organization country code                                        | `CO`                                           |
+| `SELF_ISSUED_VTC_ORG_REGISTRYID`         | Organization registry ID                                         | `1234567890`                                   |
+| `SELF_ISSUED_VTC_ORG_REGISTRYURL`        | Organization registry URL                                        | `https://registro-empresas.ejemplo.com`        |
+| `SELF_ISSUED_VTC_ORG_ADDRESS`            | Organization address                                             | `Calle Falsa 123, Bogotá, Colombia`            |
+| `SELF_ISSUED_VTC_SERVICE_TYPE`           | Service type for example credential                             | `HealthCheckService`                           |
+| `SELF_ISSUED_VTC_SERVICE_DESCRIPTION`    | Service description                                              | `Servicio de verificación de salud digital`     |
+| `SELF_ISSUED_VTC_SERVICE_MINIMUMAGEREQUIRED` | Minimum age required for service                              | `18`                                           |
+| `SELF_ISSUED_VTC_SERVICE_TERMSANDCONDITIONS` | Terms and conditions URL                                     | `https://servicio.ejemplo.com/terminos`        |
+| `SELF_ISSUED_VTC_SERVICE_PRIVACYPOLICY`  | Privacy policy URL                                               | `https://servicio.ejemplo.com/privacidad`      |
+
+> **Note:**  
+> This Self-Verifiable Trust Registry API and its configuration are **unstable** and intended for testing and development only. These endpoints and related environment variables may be removed or changed in future releases **without prior notice**.
+>
+> The variables `AGENT_LABEL` and `AGENT_INVITATION_IMAGE_URL` will be used as the name and logo for services and credentials issued by the Self-Verifiable Trust Registry.
+
+For **more examples of how to configure these variables and use the API**, see the additional README [here](../../doc/self-tr-routes.md).
 
 ## Deploy and run
 
--- a/doc/self-tr-routes.md
+++ b/doc/self-tr-routes.md
@@ -20,30 +20,6 @@ addSelfTrRoutes(app, agent, publicApiBaseUrl)
 
 ---
 
-## Configuration
-
-To enable the Self-Verifiable Trust Registry API endpoints, you must set the following environment variables in your `.env` file or system environment. These variables control the agent's identity, endpoints, and the data used for example credentials:
-
-| Variable                        | Description                                                      | Example Value                                  |
-|----------------------------------|------------------------------------------------------------------|------------------------------------------------|
-| `SELF_ISSUED_VTC_ORG_TYPE`               | Organization type for example credential                         | `PRIVATE`                                      |
-| `SELF_ISSUED_VTC_ORG_COUNTRYCODE`        | Organization country code                                        | `CO`                                           |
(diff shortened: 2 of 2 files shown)


apps/vs-agent/README.md
```diff
@@ -112,7 +112,7 @@ VS-A fetches capabilities from the `discovery.json` file (which is located at at
 
 ### Self VTR
 
-If the `SELF_VTR_ENABLED` environment variable is set to `true`, VS-A will enable features related to the Verifiable Trust Registry ([Self VTR Configuration Guide](./doc/self-vtr-routes.md)).
+If the `TESTVTR_ENABLE` environment variable is set to `true`, VS-A will enable features related to the Verifiable Trust Registry ([Self VTR Configuration Guide](./doc/self-vtr-routes.md)).
```
genaris, line 115: Add some explanation about the variables and their default value. And a note regarding the org/service name and logo that will be the ones from AGENT_LABEL and AGENT_INVITATION_IMAGE_URL respectively

apps/vs-agent/src/config/constants.ts
```diff
@@ -72,5 +72,21 @@ export const USE_CORS = Boolean(process.env.USE_CORS || false)
 export const AGENT_INVITATION_BASE_URL = process.env.AGENT_INVITATION_BASE_URL ?? 'https://hologram.zone/'
 export const REDIRECT_DEFAULT_URL_TO_INVITATION_URL =
   process.env.REDIRECT_DEFAULT_URL_TO_INVITATION_URL !== 'false'
-export const SELF_VTR_ENABLED = process.env.SELF_VTR_ENABLED === 'true'
+export const TESTVTR_ENABLE = process.env.TESTVTR_ENABLE === 'true'
```
genaris, line 75: As discussed, for the moment we can avoid using this flag. VS-A will always include Linked VPs wit self-issued credentials, unless "third-party-issued" ones are uploaded (out of scope for 1.4.0).

apps/vs-agent/src/config/constants.ts
```diff
@@ -72,5 +72,21 @@ export const USE_CORS = Boolean(process.env.USE_CORS || false)
...
 export const USER_PROFILE_AUTODISCLOSE = process.env.USER_PROFILE_AUTODISCLOSE === 'true'
+
+// Values for Organization credential
+export const TESTVTR_ORG_TYPE = process.env.TESTVTR_ORG_TYPE ?? 'PUBLIC'
```
genaris, line 79: What do you think about using SELF_ISSUED_VTC_ as prefix?

E.g.:
SELF_ISSUED_VTC_ORG_TYPE, SELF_ISSUED_VTC_ORG_COUNTRYCODE, SELF_ISSUED_VTC_SERVICE_TYPE, SELF_ISSUED_VTC_SERVICE_DESCRIPTION?

apps/vs-agent/src/testVtrRoutes.ts
```diff
@@ -32,61 +32,78 @@ import {
...
 
 // Main function to add all test routes to the Express app
-export const addSelfVtrRoutes = async (app: express.Express, agent: VsAgent, publicApiBaseUrl: string) => {
+export const addTestVtrRoutes = async (app: express.Express, agent: VsAgent, publicApiBaseUrl: string) => {
   // Create a Verifiable Presentation for ECS Service
   // Verifiable JsonSchemaCredential
   // Register endpoints for example Verifiable Presentations
   await registerVerifiablePresentationEndpoint(
-    '/self-vtr/ecs-service-c-vp.json',
+    '/test-vtr/ecs-service-c-vp.json',
```
genaris, line 70:
```suggestion
    '/self-vt/ecs-service-c-vp.json',
```
We have objects belonging to our "self trust registry" and "self verifiable public registry". Maybe it is enough to englobe all of them into a "self verifiable trust" namespace?

Or should we use 'self-tr' for JSON schema credentials and 'self-vpr' for the root JSON Schemas?

apps/vs-agent/src/config/constants.ts
lotharking: i leke it

apps/vs-agent/src/config/constants.ts
lotharking: At the moment, are we going to define a new upload method, or should we keep using the old one?

lotharking: created https://github.com/2060-io/vs-agent/issues/194for upload method

apps/vs-agent/src/utils/data.ts
```diff
@@ -0,0 +1,236 @@
+import { PUBLIC_API_BASE_URL } from '../config'
```
genaris, line 1: This will not work, since now we are encouraging users to only define AGENT_PUBLIC_DID. I think we must find a different way to render these schemas. Probably turning this into a function, so the publicApiBaseUrl is passed from the controller (which has access to Agent instance).

apps/vs-agent/README.md
```diff
@@ -110,9 +109,9 @@ When connecting to other agents, VS-A tries to get information from them in orde
 
 VS-A fetches capabilities from the `discovery.json` file (which is located at at `/www/apps/vs-agent/discovery.json` in the deployed container) to determine available features. If you want to customize the capabilities to look for, replace the volume at this path with your own `discovery.json` file.
 
-### Self VTR
+### Self VR
```
genaris, line 112: I think we must refactor this README to include self tr environment variables as one of the subections of the main VS configuration.

We can also keep the details of the implementation in doc/self-tr-routes and add a link here, but since 99% of the users will be interested only in the config, I don't see why we'll force them to use two documents.

doc/self-tr-routes.md
```diff
@@ -0,0 +1,91 @@
...
+| `SELF_ISSUED_VTC_SERVICE_TYPE`           | Service type for example credential                             | `HealthCheckService`                           |
+| `SELF_ISSUED_VTC_SERVICE_DESCRIPTION`    | Service description                                              | `Servicio de verificación de salud digital`     |
+| `SELF_ISSUED_VTC_SERVICE_MINIMUMAGEREQUIRED` | Minimum age required for service                              | `18`                                           |
+| `SELF_ISSUED_VTC_SERVICE_TERMSANDCONDITIONS` | Terms and conditions URL                                     | `https://servicio.ejemplo.com/terminos`        |
+| `SELF_ISSUED_VTC_SERVICE_PRIVACYPOLICY`  | Privacy policy URL                                               | `https://servicio.ejemplo.com/privacidad`      |
+
+> **Note:**  
+> This Self-Verifiable Trust Registry API and its configuration are **provisional** and intended for testing and development only. These endpoints and related environment variables may be removed or changed in future releases **without prior notice**.
```
genaris, line 41: It is unstable, but not meant to be provisional.

genaris (changes requested):
Some minor comments regarding document organization. And an important one about schemas base URL.

PR #208, opened 2025-08-15
fix: update permission response

lotharking:
updated based on spec


genaris (commented):
Where is this response defined in the spec and where are we using it? For instance, what's the purpose of returning the same did we are querying for?

PR #212, opened 2025-08-28
feat: support DIDComm and self-signed Verifiable Trust on did:webvh

genaris:
Current code base only allows to create did:webvh but not in an useful way, since it does not work in DIDComm due to verification key issues and it does not add self signed verifiable trust Linked VPs for it.

Here there is a major refactor in the way Agent is initialized, simplifying `setupVsAgent` and leaving most of the specific work to the `VsAgent` class.

Besides some fixes on WebVhDidRegistrar, we add also a plain WebDidRegistrar that might be useful for others so we can maybe create a PR for credo-ts core as well.

There is a patch applied to didwebvh-ts to solve a problem with DID update. There are more fixes to be done in that library, but related to client side so by now it is affecting only Hologram.

### DIFF
--- a/apps/vs-agent/package.json
+++ b/apps/vs-agent/package.json
@@ -57,7 +57,7 @@
   "dependencies": {
     "@2060.io/credo-ts-didcomm-calls": "^0.0.10",
     "@2060.io/credo-ts-didcomm-media-sharing": "0.0.2",
-    "@2060.io/credo-ts-didcomm-mrtd": "0.0.14",
+    "@2060.io/credo-ts-didcomm-mrtd": "0.0.15",
     "@2060.io/credo-ts-didcomm-receipts": "0.0.7",
     "@2060.io/credo-ts-didcomm-user-profile": "0.0.6",
     "@2060.io/vs-agent-model": "workspace:*",
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -91,8 +91,8 @@ importers:
         specifier: 0.0.2
         version: 0.0.2(expo@53.0.9(@babel/core@7.27.3)(react-native@0.79.2(@babel/core@7.27.3)(react@19.1.0))(react@19.1.0))(react-native@0.79.2(@babel/core@7.27.3)(react@19.1.0))(web-streams-polyfill@3.3.3)
       '@2060.io/credo-ts-didcomm-mrtd':
-        specifier: 0.0.14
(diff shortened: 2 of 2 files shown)


apps/vs-agent/src/utils/WebVhDidRegistrar.ts
```diff
@@ -101,10 +178,9 @@ export class WebVhDidRegistrar implements DidRegistrar {
         keyAgreement: normalizeMethodArray(keyAgreement),
         services,
       })
-
-      const didDocument = new DidDocument(doc)
+      console.log(`logResult: ${JSON.stringify(logResult)}`)
```
genaris, line 181:
```suggestion
```

apps/vs-agent/src/utils/WebVhDidRegistrar.ts
```diff
@@ -55,15 +121,26 @@ export class WebVhDidRegistrar implements DidRegistrar {
    */
   public async update(agentContext: AgentContext, options: WebVhDidUpdateOptions): Promise<DidUpdateResult> {
     try {
-      const { did, domain, didDocument: inputDidDocument } = options
+      const { did, didDocument: inputDidDocument } = options
```
lotharking, line 124: Why not just call it `didDocument`? The `inputDidDocument` was overridden because it was used to identify the `didDocument` input object. The `didDocument` variable should be more consistent with the result of the create method
```suggestion
      const { did, didDocument } = options
```

apps/vs-agent/src/utils/VsAgent.ts
```diff
@@ -63,13 +74,223 @@ interface AgentOptions<VsAgentModules> {
...
+      displayName: this.config.label,
+      displayPicture,
+    })
+
+    const parsedDid = this.did ? parseDid(this.did) : null
+    if (parsedDid) {
+      // If a public did is specified, check if it's already stored in the wallet. If it's not the case,
+      // create a new one and generate keys for DIDComm (if there are endpoints configured)
+      // TODO: Make DIDComm version, keys, etc. configurable. Keys can also be imported
+      const domain = parsedDid.id.includes(':') ? parsedDid.id.split(':')[1] : parsedDid.id
```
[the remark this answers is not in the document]
lotharking: No issues. I get it now. It’s okay for me.

lotharking (approved):

PR #226, opened 2025-09-09
feat: add deployment requests & limits + docs

gabrielmatau79:
## Summary
 
- Add resources for main vs-agent container, redis and postgres in values.yaml and deployment.yaml
- Update chart documentation with new Resources section and example values

**note:** The values ​​for the requests and limits were taken from the implementation that is in the Hologram backend of these components (redis, postgres).


lotharking:
Hey @gabrielmatau79, don't forget to update the chatbot example too, and when generating the new version, test the update on demos deploy the chatbot demo

PR #254, opened 2025-10-21
fix: credential key generation logic and improve naming clarity

lotharking:
This PR refactors the credential key generation method to improve naming clarity, validation, and documentation, while adding a `@todo` to review whether this is the best approach for storing keys in `GenericRecord`. It also removes the destructuring statement `const { proof: _proof, ...rest } = (didRecord.metadata.get(id) as W3cJsonLdVerifiableCredential) ?? null` to prevent runtime errors when the value is null or undefined.


lotharking:
Hey @genaris,
Created issue https://github.com/2060-io/vs-agent/issues/255 based on this implementation for improving the logic in this `TrustService`

PR #261, opened 2025-10-24
fix: remove duplicate context at did creation

lotharking:
#### Context creation
When the DID is created, the resulting context looks like this:

```json
{
  "@context": [
    "https://www.w3.org/ns/did/v1",
    "https://w3id.org/security/multikey/v1",
    "https://w3id.org/security/multikey/v1",
    "https://w3id.org/security/suites/ed25519-2020/v1",
    "https://w3id.org/security/suites/x25519-2019/v1",
    "https://identity.foundation/linked-vp/contexts/v1"
  ],
  ...
}
```
As you can see, some contexts are duplicated.
That’s why it’s necessary to add a setter control to prevent duplicate entries.

---
#### Upload credential
Update the DID Document while preventing duplicate entries

### DIFF
--- a/apps/vs-agent/src/controllers/admin/verifiable/TrustService.ts
+++ b/apps/vs-agent/src/controllers/admin/verifiable/TrustService.ts
@@ -103,10 +103,9 @@ export class TrustService {
         type: 'LinkedVerifiablePresentation',
       })
 
-      // TODO: review if possible allow duplicate id or endpoint
       if (didRecord.didDocument) {
         didRecord.didDocument.service = (didRecord.didDocument?.service ?? []).filter(
-          s => s.id !== newService.id && s.serviceEndpoint !== newService.serviceEndpoint,
+          s => s.id !== newService.id,
         )
         didRecord.didDocument?.service.push(newService)
       }


apps/vs-agent/src/controllers/admin/verifiable/TrustService.ts
```diff
@@ -97,13 +97,19 @@ export class TrustService {
...
+      const newService = new DidDocumentService({
+        id: `${agent.did}#vpr-${schemaKey}-c-vp`,
+        serviceEndpoint,
+        type: 'LinkedVerifiablePresentation',
+      })
+
+      // TODO: review if possible allow duplicate id or endpoint
+      if (didRecord.didDocument) {
+        didRecord.didDocument.service = (didRecord.didDocument?.service ?? []).filter(
+          s => s.id !== newService.id && s.serviceEndpoint !== newService.serviceEndpoint,
```
genaris: I don't get why you are looking also by serviceEndpoint. This makes it possible to have two services with the same id and that would be an issue since it is not compatible with the DID Core spec and may result in issues when resolving the DID Document:

> The value of the id property MUST be a [URI](https://www.w3.org/TR/did-1.0/#dfn-uri) conforming to [[RFC3986](https://www.w3.org/TR/did-1.0/#bib-rfc3986)]. A [conforming producer](https://www.w3.org/TR/did-1.0/#dfn-conforming-producer) MUST NOT produce multiple service entries with the same id. A [conforming consumer](https://www.w3.org/TR/did-1.0/#dfn-conforming-consumer) MUST produce an error if it detects multiple service entries with the same id.

genaris (changes requested):

PR #275, opened 2025-11-18
fix: allow querying anoncreds resources by object type

lotharking:
**Changes:**
* Updated the `saveAttestedResource` method to allow including `resourceType` in `genericRecords`.
* Updated the delete control to use an API query parameter for stronger security, since passing the ID as a path parameter was causing issues.

Hey @genaris, I prefer using `resourceType` values like `anonCredsSchema`, `anonCredsCredDef`, `anonCredsRevocRegDef`, and `anonCredsStatusList` as the metadata resource type registries based on [examples](https://github.com/openwallet-foundation/credo-ts/blob/main/packages/webvh/src/anoncreds/services/__tests__/mock-resources.ts).


genaris (approved):
Nicely implemented!

PR #280, opened 2025-11-24
fix: swager enablement is through a configuration flag

lotharking:
**Changes:**
- Created ENABLE_SWAGGER environment.
- Added swagger enablement control
- Includes in charts control (by default it's disabled due to  testing purposes)
- Updates docs
> **Note:** Only public API


genaris (approved):
I approve but please update variable naming.

PR #303, opened 2026-01-08
feat: add related JSON Schema Credential ID to credential type creati…

genaris:
- Add an optional field to POST /credential-types to specify the related JSON Schema Credential ID
- Some refectoring: type -> format, jsonSchemaCredential -> jsonSchemaCredentialId
- Move VT related endpoints: 
   - /agent/vtc -> /vt/linked-credentials
   - /agent/jsc -> /vt/json-schema-credentials

### DIFF
--- a/apps/vs-agent/tests/trustService.test.ts
+++ b/apps/vs-agent/tests/trustService.test.ts
@@ -167,9 +167,7 @@ describe('TrustService', () => {
         expect.objectContaining({
           status: 200,
           didcommInvitationUrl: expect.any(String),
-          credential: expect.objectContaining({
-            credentialExchangeId: expect.any(String),
-          }),
+          didcommCredentialExchangeId: expect.any(String),
         }),
       )
     }, 20000)


apps/vs-agent/src/controllers/admin/credentials/CredentialTypeController.ts
```diff
@@ -90,7 +90,7 @@ export class CredentialTypesController {
           version: (record.getTag('version') as string) ?? schema?.version,
           attributes: schema?.attrNames || [],
           revocationSupported,
-          relatedJsonSchemaCredential: record.getTag('relatedJsonSchemaCredential') as string,
+          relatedJsonSchemaCredentialId: record.getTag('relatedJsonSchemaCredentialId') as string,
```
lotharking: In the Post method example, it would be great to include it in the examples.

genaris: Ok I think this will be part of the following PR including improvements to documentation

PR #310, opened 2026-01-08
fix: implicit invitation to legacy did:web won't autorespond

genaris:
Something that was missing from previous fix


apps/vs-agent/src/events/ConnectionEvents.ts
```diff
@@ -19,6 +19,10 @@ import { PresentationStatus, sendPresentationCallbackEvent } from './CallbackEve
 import { sendWebhookEvent } from './WebhookEvent'
 
 export const connectionEvents = async (agent: VsAgent, config: ServerConfig) => {
+  // Get the first recordm atching agent's DID and obtain all alternatives for it
```
lotharking:
```suggestion
  // Get the first record matching agent's DID and obtain all alternatives for it
```

lotharking (approved):

PR #317, opened 2026-01-16
chore: upgrade to credo 060

lotharking:


apps/vs-agent/src/utils/VsAgent.ts
```diff
@@ -95,11 +110,11 @@ export class VsAgent extends Agent<VsAgentModules> {
     await super.initialize()
 
     // Make sure default User Profile corresponds to settings in environment variables
-    const imageUrl = this.config.connectionImageUrl
+    const imageUrl = AGENT_INVITATION_IMAGE_URL
```
genaris: Rather than calling the constants from this class, why don't you keep both imageUrl and label as creation parameters and store them as plain variables of the VsAgent class (like we did for did and publicApiBaseUrl?

PR #318, opened 2026-01-20
refactor: moving to @verana-labs scope

genaris:
Updated all packages and apps to use @verana-labs scope instead of @2060.io.

Several references in docs and code were updated too. We'll still need to update some examples since in some cases verana-labs URLs are more appropriate, but that could be left to a further PR where we update other aspects of the documentation.


charts/chatbot/Chart.yaml
lotharking: It's required update the chart name for deployment

PR #334, opened 2026-02-06
fix: by default support revocation

lotharking:
**Changes**
At some point, we discontinued the use of revocation, and this feature is supported and should be adopted in the future.


genaris (commented):
I don't think it is the right moment to do this update, because revocation stuff is not yet completely polished and this could generate confusion in the demos and examples we have in the documentation.

For instance, /invitation/credential-offer POST endpoint does not include parameters related to revocation, so we'd need to add them and explain how they should be set.

So for now I'd leave this as an "advanced/experimental" feature and therefore I'd keep it as false by default.

PR #344, opened 2026-02-16
fix: include devnet on urlMap

lotharking:


mjfelis:
@genaris @lotharking can we generate a release as it is a blocking topic

PR #355, opened 2026-02-23
style: apply linter rules and fix CI script

lotharking:
**Changes**
- Review linter correlation


genaris (commented):
Is this lint command actually running in CI? I cannot find it in the logs.

PR #389, opened 2026-04-08
ci: solve deployment to npm bug present

lotharking:


genaris (approved):
Still not sure if node:24 includes something newer than npm 11.5, but let's try.

PR #392, opened 2026-04-09
fix: use a fix node version for deployment

lotharking:
At the moment, there is a bug in the latest version. (https://github.com/nodejs/node/issues/62425 and https://github.com/npm/cli/issues/9151)


genaris (approved):

genaris:
After releasing v1.10.0, I think we need to upgrade to Credo 0.7, which removes dependency on ffi-napi and ref-napi and allows us to use node 24 and make the Dockerfile building process a bit easier.

PR #394, opened 2026-04-10
fix: use related json schema presentation request

lotharking:
With this implementation, we should request the issuer to provide a proof request invitation.

### DIFF
--- a/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeService.ts
+++ b/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeService.ts
@@ -324,7 +324,7 @@ export class CredentialTypesService {
     return credentialDefinitionRecord
   }
 
-  private async fetchJson<T>(url: string): Promise<T> {
+  public async fetchJson<T>(url: string): Promise<T> {
     const res = await fetch(url)
     if (!res.ok) {
       throw new Error(`Failed to fetch ${url}: ${res.statusText}`)
--- a/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeService.ts
+++ b/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeService.ts
@@ -8,7 +8,7 @@ import { Inject, Logger } from '@nestjs/common'
 import { mapToEcosystem } from '@verana-labs/vs-agent-model'
 
 import { VsAgentService } from '../../../services/VsAgentService'
-import { VsAgent } from '../../../utils'
+import { fetchJson, VsAgent } from '../../../utils'
 
 type Tags = TagsBase & {
   type?: never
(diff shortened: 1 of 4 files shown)


apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -153,6 +153,25 @@ export class InvitationController {
           ],
         },
       },
+      withRelatedJsonSchema: {
+        summary: 'Using relatedJsonSchemaCredentialId',
+        value: {
+          requestedCredentials: [
+            {
+              relatedJsonSchemaCredentialId:
```
genaris:
```suggestion
              jsonSchemaCredentialId:
```

The term `related` is something we use to correlate credential definitions with the JSC, but in this case we just want to, at a high-level, request credentials based on a certain JSC

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -175,14 +194,37 @@ export class InvitationController {
...
-    if (attributes && !Array.isArray(attributes)) {
+    if (!rawCredentialDefinitionId && !relatedJsonSchemaCredentialId) {
+      throw new Error('Either credentialDefinitionId or relatedJsonSchemaCredentialId must be provided')
+    }
+    let credentialDefinitionId: string
+
+    if (relatedJsonSchemaCredentialId) {
+      const [credDefRecord] = await agent.modules.anoncreds.getCreatedCredentialDefinitions({
+        relatedJsonSchemaCredentialId,
+      })
```
genaris: Why are you looking at created credential definitions?

We are the verifier, not necessarily (neither usually) the ones who created the credential definition. 

When we receive a relatedJsonSchemaCredentialId, we need to retrieve the JSC, get the DID of its issuer, and if it is a did:webvh, query for the anonCredsSchema resource whose relatedJsonSchemaCredentialId is the one we are looking for.

Then, when building the proof request, instead of using cred_def_id as the restriction, we must use schema_id and use the ID of the AnonCreds Schema we retrieved from JSC issuer. If not found (e.g. because the JSC issuer wasn't using VS Agent 1.10.0 or newer), throw error.

I don't know why but I thought this was already implemented. Maybe I'm confused with the credential issuance flow, where we are also retrieving the AnonCreds Schema, but for another purpose.

genaris (changes requested):
Please verify carefully, since there are conceptual errors in this PR.

apps/vs-agent/src/controllers/admin/credentials/CredentialTypeService.ts
```diff
@@ -324,7 +324,7 @@ export class CredentialTypesService {
     return credentialDefinitionRecord
   }
 
-  private async fetchJson<T>(url: string): Promise<T> {
+  public async fetchJson<T>(url: string): Promise<T> {
```
genaris: I think this can be moved to an utility module rather than this class

genaris (approved):

PR #396, opened 2026-04-14
chore(deps): upgrade credo-ts to v0.7.0-alpha-20260410113934

### LINKED ISSUE #393: Upgrade to Credo 0.7.0
The upcoming credo-ts 0.7.0 will upgrade some of its dependencies, notably AnonCreds and Askar JavaScript wrappers, which do not rely on ffi-napi anymore. This means that the setup would be easier, since we won't need to build the native modules every time. Although I don't think we can use an alpine-based image, at least I think we can simplify the image building process since things like python won't be needed anymore (AFAIK).

For the moment, we can test with the latest credo-ts 0.7.0 alpha and see how well it goes for us.

tarunvadde:
Upgrade all @credo-ts/* packages from v0.6.3 to v0.7.0-alpha, along with
native bindings that now use koffi instead of ffi-napi, eliminating the
need for Python/make/g++ during builds.

- Upgrade @credo-ts/{core,didcomm,askar,anoncreds,node,action-menu,
  question-answer,webvh} to 0.7.0-alpha-20260410113934
- Upgrade @hyperledger/anoncreds-nodejs 0.2.4 => 0.4.0
- Upgrade @openwallet-foundation/askar-nodejs 0.4.3 => 0.6.0
- Upgrade @2060.io companion modules to 0.1.x
- Upgrade credo-ts-didweb-anoncreds to 0.0.2-alpha.4
- Adapt LogLevel enum to PascalCase (warn => Warn, debug => Debug, etc.)
- Adapt 2060.io/user-profile renames (DidComm* prefix on types/events)
- Replace JsonEncoder.fromBuffer with JsonEncoder.fromUint8Array
- Remove @credo-ts/action-menu patch (fix included in 0.7.0)
- Patch credo-ts-didweb-anoncreds exports for CJS compatibility
- Simplify Dockerfile: remove python3/make/g++, use node:22.19-slim

Closes #393


apps/vs-agent/src/utils/logger.ts
```diff
@@ -9,13 +9,13 @@ export class TsLogger extends BaseLogger {
...
   // Map our log levels to tslog levels
   private tsLogLevelMap = {
-    [LogLevel.test]: 'debug',
-    [LogLevel.trace]: 'debug',
-    [LogLevel.debug]: 'debug',
-    [LogLevel.info]: 'log',
-    [LogLevel.warn]: 'warn',
-    [LogLevel.error]: 'error',
-    [LogLevel.fatal]: 'fatal',
+    [LogLevel.Test]: 'debug',
```
lotharking, line 12: Is this required by credo 0.7.0? If not, I prefer to create another PR to avoid misunderstanding.

tarunvadde: Hi @lotharking,

As per the recent changes for credo 0.7.0, this is required. You can refer to the original PR for this.

https://github.com/openwallet-foundation/credo-ts/pull/2724

patches/credo-ts-didweb-anoncreds@0.0.2-alpha.4.patch
genaris, line 1: Try with the recently released [v0.0.2-alpha.5](https://github.com/2060-io/credo-ts-didweb-anoncreds/releases/tag/v0.0.2-alpha.5).

genaris (changes requested):

genaris (approved):
Thanks! I've added a .gitkeep file for patches directory now that it is empty to prevent issues on docker image generation. Also updated a bit examples Dockerfiles because they were failing.

PR #399, opened 2026-04-16
feat: support reactions module

### LINKED ISSUE #398: [Reactions] Support reaction messages in Vs-agent
It looks like they aren't supported yet but are really helpful for AI agent sessions.

Both set by user and service.

lotharking:
**Changes**
- Include the reactions module


apps/vs-agent/src/controllers/admin/message/MessageService.ts
```diff
@@ -98,6 +99,17 @@ export class MessageService {
...
           receipts: textMsg.receipts.map(didcommReceiptFromVsAgentReceipt),
         })
+      } else if (messageType === ReactionMessage.type) {
+        const msg = JsonTransformer.fromJSON(message, ReactionMessage)
+        await agent.modules.reactions.send({
+          connectionId: msg.connectionId,
+          reactions: msg.reactions.map(r => ({
+            messageId: r.messageId,
+            emoji: r.emoji,
+            action: r.action as any,
```
tarunvadde: Can we eliminate the use of `any`?

Maybe by using a real type: `export type ReactionAction = 'react' | 'unreact'`, then validate with `@IsIn(['react','unreact'])` on the class property.

packages/model/src/messages/ReactionMessage.ts
```diff
@@ -0,0 +1,71 @@
...
+  reactions: VsAgentMessageReactionOptions[]
+}
+
+export class ReactionMessage extends BaseMessage {
+  public constructor(options: ReactionMessageOptions) {
+    super()
+
+    if (options) {
+      this.id = options.id ?? this.generateId()
+      this.threadId = options.threadId
```
tarunvadde: Are we not passing `threadId` to `agent.modules.reactions.send(...)`?

lotharking: This protocol doesn't have a threadID

genaris (approved):

tarunvadde (approved):

PR #401, opened 2026-04-16
feat: use verana indexer websocket

lotharking:
**Changes**
* Start with v4 specification, including a listener for the indexer WebSocket
* Create `IndexerWebSocketService.ts` to:
  * establish a persistent WebSocket connection filtered by DID
  * handle reconnection with exponential backoff
  * perform re-synchronization via REST after reconnect
  * buffer WebSocket events during initial sync
  * persist and track `lastBlockHeight`
  * maintain local state cache for Trust Registries, Credential Schemas, and Permissions
* Create `VeranaIndexerService.ts` to manage HTTP requests to the indexer (used for sync and recovery)
* Update docs and types

> Note:
> This PR covers the transport and synchronization aspects of §7 (Indexer WebSocket Subscription).
>
> The following are intentionally out of scope and will be handled in separate PRs:
> * §6 — Automatic VTJSC generation and DID Document publication
> * §7 — Notification handler architecture and side effects (e.g., VTJSC publication, flow progression, credential cleanup) https://github.com/verana-labs/vs-agent/pull/422
>
> Current implementation focuses on state mirroring (`applyActivity`) and does not yet include dispatch-based handlers or side-effect execution.


genaris (commented):
We didn't define yet the protocol we are going to use to notify changes from indexer to VS Agent. I don't think it makes any sense to keep implementing things blindly.

Leave this as draft for now.

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
```diff
@@ -0,0 +1,199 @@
...
+
+    ws.on('open', () => {
+      this.logger.info(`[IndexerWS] Connected to indexer`)
+      this.reconnectAttempt = 0
+    })
+
+    ws.on('message', (data: WebSocket.RawData) => {
+      try {
+        const event = JSON.parse(data.toString()) as JsonObject
+        const blockHeight = (event['block_height'] as number | undefined) ?? 0
```
tarunvadde, line 87: Hi Andres, a question on the bootstrap sequencing. VPR Integration Spec §2.2 step 3 says:

> "Subscription returns a `block-height`. Query the indexer for all objects linked to the agent's DID at `block-height - 1` to initialize the agent's state. Start processing WebSocket notifications from `block-height.`"

The indexer's welcome message carries this cursor (from `verana-indexer/src/services/api/events_broadcaster.ts)`:

```
const connectionMessage = {
  type: "connected",
  block_height: detailedStatus?.lastProcessedBlock ?? null,
  ...
};
```

The current message handler reads `event['block_height']` then exits at the `payload?.sender` check (welcome message has no payload), so the welcome cursor isn't captured. The REST sync currently uses `loadSyncState().lastBlockHeight` instead.

Is the welcome `block_height` supposed to be the cursor the spec refers to? If yes, capturing it (e.g. by branching on `event.type === 'connected'` before the payload check) would tie the bootstrap to the spec's wording. If you intentionally went with persisted state, happy to hear the reasoning.

packages/agent-sdk/src/blockchain/VeranaIndexerService.ts
```diff
@@ -0,0 +1,28 @@
...
+    const url = `${this.baseUrl}/verana/indexer/v1/changes/${blockHeight}`
+    const data = await fetchJson<Record<string, unknown>>(url)
+    return {
+      block_height: Number(data['block_height'] ?? blockHeight),
+      next_change_at: data['next_change_at'] != null ? Number(data['next_change_at']) : null,
+      activity: (data['activity'] ?? []) as IndexerActivity[],
+    }
+  }
+
+  async getEvents(agentDid: string, afterBlockHeight = 0, limit = 100): Promise<IndexerEventsResponse> {
```
tarunvadde: A question on the catch-up loop. The indexer's `/events` endpoint has `default=100, max=500` and the response's `after_block_height` is the input echoed back (`verified in verana-indexer/src/services/api/events.service.ts`):

`return { events, count: events.length, after_block_height: afterBlockHeight };`

`getEvents` here is a single call with `limit=100` and no loop. If the agent has more than 100 events to catch up on, the remaining events aren't fetched - the next call would only happen on the next reconnect, with the same `after_block_height` if no relevant blocks have advanced `lastBlockHeight` in the meantime.

Was the intent to handle catch-up across multiple pages? If yes, a loop that tracks the last `block_height` in `events[]` as the next call's `after_block_height` would do it. Worth bumping the limit to 500 (the indexer's max) too.

tarunvadde (commented):
LGTM. Some clarifications on the questions posted would be helpful for me to understand this better.

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
lotharking: Good point @tarunvadde, thanks for the review.
You're right, the `connected` message returns the `lastProcessedBlock`, but I was using the local value as the starting point. Since §2.2 mentions fetching events from a given block height, and the `/events` endpoint only needs that starting point, I based it on the persisted value.

packages/agent-sdk/src/blockchain/VeranaIndexerService.ts
lotharking: Yes, you're right.
Right now it's a single call. I'll use the max limit (500), but pagination should be handled on the Indexer side to cover all events.

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
lotharking: Hey @mabubakr1113, currently, this part of the WebSocket returns the last block processed. However, I don’t have an easy way to get the current state of all past objects related to this DID.
At the moment, I’m using the `/events` endpoint, but it reconstructs state from events rather than returning the actual objects.
I’ve also created an issue to track simplifying the initial sync required by §2.2. https://github.com/verana-labs/verana-indexer/issues/232

apps/vs-agent/src/main.ts
```diff
@@ -237,6 +239,19 @@ const run = async () => {
     plugin.registerEvents?.(agent, conf)
   }
 
+  // Connect to Verana indexer for on-chain notifications
+  if (VERANA_INDEXER) {
```
tarunvadde: is VERANA_INDEXER optional? Because our spec says it's REQUIRED

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
```diff
@@ -0,0 +1,204 @@
...
+        if (blockHeight <= 0) return
+
+        const payload = event['payload'] as { sender?: string } | undefined
+        if (!payload?.sender) return
+        const sender = payload.sender
+
+        if (this.syncing) {
+          this.pendingEvents.push({ blockHeight, sender })
+        } else {
+          this.processBlock(blockHeight, sender).catch(err =>
```
tarunvadde: Can we serialise live event processing the same way drain does - push into a queue with an awaiting consumer, or chain promises?

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
```diff
@@ -0,0 +1,204 @@
...
+  private reconnectAttempt = 0
+  private stopped = false
+  private syncing = false
+  private pendingEvents: PendingEvent[] = []
+  private readonly indexer: VeranaIndexerService
+
+  constructor(private readonly options: IndexerWebSocketServiceOptions) {
+    this.indexer = new VeranaIndexerService({
+      baseUrl: options.indexerUrl,
+      logger: options.agent.config.logger as never,
```
tarunvadde: Can we instantiate this with a proper type Credo uses such as BaseLogger or Logger?

apps/vs-agent/src/config/constants.ts
```diff
@@ -117,6 +117,7 @@ export const AGENT_AUTO_UPDATE_STORAGE_ON_STARTUP =
 export const AGENT_BACKUP_BEFORE_STORAGE_UPDATE = process.env.AGENT_BACKUP_BEFORE_STORAGE_UPDATE !== 'false' // removed on credo-ts v0.6.0
 
 // Verana network
+export const VERANA_INDEXER = process.env.VERANA_INDEXER
```
genaris: Verana indexer what? 
VERANA_INDEXER_URL or VERANA_INDEXER_BASE_URL would be consistent with other settings.

apps/vs-agent/src/main.ts
genaris: It is required to use Verana VPR integration features, but currently we make the whole feature optional (as indicated in the README). Probably it will become mandatory and made this a core feature of VS Agent as soon as everything is in an stable state.

apps/vs-agent/src/main.ts
```diff
@@ -237,6 +239,19 @@ const run = async () => {
...
+    try {
+      const indexerWs = new IndexerWebSocketService({
+        indexerUrl: VERANA_INDEXER,
+        agent,
+      })
+      await indexerWs.start()
+    } catch (error) {
+      serverLogger.error(`Failed to start Verana indexer connection: ${(error as Error).message}`)
```
genaris: Do you think it is ok to catch the error? I'm wondering that probably if somebody expliclitly sets the Verana feature flags is because they expect the agent to connect to it. If it fails, shouldn't the app exit and the container restart?

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
```diff
@@ -0,0 +1,204 @@
...
+    this.ws = null
+  }
+
+  private get logger() {
+    return this.options.agent.config.logger
+  }
+
+  private get agentDid(): string {
+    return this.options.agent.did ?? ''
+  }
```
genaris: Does this make sense if agent did is undefined? I think we shouldn't even start this service in case it is not defined. Or at least throw an error when attempting to read agentDid

genaris (changes requested):
Some minor comments regarding config.

apps/vs-agent/src/main.ts
tarunvadde: Got it

packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
lotharking: IMHO, I don’t think this is required. Normally, every process on the ledger takes a long time (each transaction requires confirmation or similar, at least about a minute). With this in mind, I think adding this for event processing would be overengineering for this case.

genaris (approved):
Let's merge this so we can progress on the handling of those events so we can start integrating the vt-flow protocol and see how agents can interact.

PR #410, opened 2026-04-20
feat: split docker image to one based on plugins

### LINKED ISSUE #408: Remove MRTD dependencies from base vs-agent image
One of the remaining work from #383 is to find a way to not force all VS Agents to use MRTD if they don't need so. The same applies for chat, although that one is not that critical because it doesn't add any significant dependency to the container.

An initial idea is to create a base image that doesn't include it, and some extended images that simply add the plugins. They will be published as "flavours", e.g. vs-agent-chat and vs-agent-mrtd or something (more aligned with standard conventions).

lotharking:
Create 3 different images, each one based on its respective plugins.
**Changes**
- Update cd, Dockerfile, and docker compose.
- Move `requestUserProfile` to chatPlugin
- Update the image of values to the deployment


.github/workflows/cd.yml
```diff
@@ -67,20 +70,80 @@ jobs:
...
           file: ./apps/vs-agent/Dockerfile
+          target: vs-agent
           push: true
           tags: |
             ${{ secrets.DOCKER_HUB_LOGIN }}/${{ env.IMAGE_NAME }}:${{ env.IMAGE_TAG }}
             ${{ secrets.DOCKER_HUB_LOGIN }}/${{ env.IMAGE_NAME }}:v${{ env.RELEASE_MAJOR }}${{ env.SUFFIX }}
             ${{ secrets.DOCKER_HUB_LOGIN }}/${{ env.IMAGE_NAME }}:v${{ env.RELEASE_MAJOR }}.${{ env.RELEASE_MINOR }}${{ env.SUFFIX }}
             ${{ secrets.DOCKER_HUB_LOGIN }}/${{ env.IMAGE_NAME }}:${{ env.RELEASE_VERSION }}
 
+      # VS Agent Chat (base + chat plugin)
```
genaris: It looks that both are doing the same thing, except that  they use a different image name. Can't we just iterate on them using a matrix or something like that?

I don't think it would be a problem to make `target` to match the image name instead of using separate variables for image names.

genaris (approved):

Issue #405, opened 2026-04-17
Create VT Flow Credo DIDComm extension


genaris:
Dependent on #404.

Implement an extension module for Credo 0.7 that supports all messages and states defined in the protocol. Make sure the API allows to manually manage every state needed, e.g. to be able to perform some blockchain commands in-between different states of the protocol.

Don't forget to add some unit test suite to make sure no regressions will be added in the future.

genaris:
Initial implementation on #411

PR #411, opened 2026-04-21
feat(vt-flow): add Credo-TS superprotocol module

### LINKED ISSUE #404: Design VT DIDComm protocol
Based on the section 5 (Validation Process and Issuance Flows), the goal is to design and specify a DIDComm protocol (following the conventions in Aries RFCs and didcomm.org) suitable for .

The protocol specification can be initially hosted in vs-agent repo, under `docs` directory, at least until we create a dedicated Credo extension package for it.

A brief summary of what has been discussed by Discord:

> We want to implement a vt_flow, which will be a superprotocol that:
> - Carries the Verana-specific state (perm_id, session_uuid, agent_perm_id, wallet_agent_perm_id, flow state machine)
> - Has its own message types (VR, IR, OOB_LINK, VALIDATING, CRED_ACCEPT, CRED_STATE_CHANGE, ERROR)
> - Delegates the actual credential delivery to the standard Issue Credential V2 protocol as a subprotocol
> - Delegates proof checking (when needed) to Present Proof V2 as a subprotocol
> 
> Issue Credential V2/Present Proof V2 subprotocol runs on its own thread, but every message carries pthid: so both agents know it's happening inside the vt_flow session.
> 

Make sure that message types follow the DIDComm conventions (kebab-case for the type and fields) and reuse other message types whenever needed: for instance, in case of errors we can probably use Problem Report.

Once we have this protocol properly defined and discussed, we'll can proceed on the implementation.

tarunvadde:
## Summary

Adds `packages/vt-flow` — a standalone Credo-TS DIDComm module implementing the Verifiable Trust Flow (vt-flow) superprotocol v1.0 (`https://didcomm.org/vt-flow/1.0`).

Closes #404, #405.

## What's in this PR

- **`packages/vt-flow/`** — new package:
  - 5 wire messages: `validation-request`, `issuance-request`, `oob-link`, `validating`, `credential-state-change`
  - Applicant and Validator roles (perspective-only, never on the wire)
  - 16 Flow States + 3 Connection States per spec §5.6
  - Credo-TS `Module` with DI registration, handler registry, and `DidCommCredentialStateChanged` subprotocol correlation via `~thread.pthid`
  - `VtFlowApi` with one method per state transition so callers can gate each on on-chain work
  - Per-flag opt-in auto-chain (`autoAcceptValidationRequest`, `autoMarkValidated`, `autoOfferCredential`, etc.) for unattended deployments
  - RFC 0035 problem-report builder with per-code `who_retries` / `impact` / `retryable` metadata
  - `VtFlowStateChangedEvent` emitted on every transition
  - 54 unit tests

- **`apps/vs-agent` — test infra only**, no production changes:
  - `@verana-labs/credo-ts-didcomm-vt-flow` added as `optionalDependency`
  - Two-agent integration test (`tests/vtFlow.test.ts`)
  - `setupVtFlow` hookup in the test mock's `startAgent` helper
  - `vitest fileParallelism: false` to serialise askar-nodejs native bindings under concurrent agents

## Test plan

- [x] `pnpm build` clean at workspace root
- [x] `packages/vt-flow` — 54 unit tests pass (enums, record, messages)
- [x] `apps/vs-agent` — 10 integration tests pass including the two-agent vt-flow round-trip (`tests/vtFlow.test.ts`) covering `thid` correlation, Validator auto-chain to `VALIDATING`/`VALIDATED`, and both-side record metadata
- [x] `pnpm lint` clean


packages/vt-flow/src/VtFlowVariant.ts
```diff
@@ -0,0 +1,6 @@
+export enum VtFlowVariant {
+  /** §5.1 — Schema management mode `GRANTOR` or `ECOSYSTEM`; opened by `validation-request`. */
+  ValidationProcess = 'validation-process',
+  /** §5.2 — Direct issuance (HOLDER/ISSUER); opened by `issuance-request`. */
+  DirectIssuance = 'direct-issuance',
+}
```
lotharking: I understand the intention behind the comments, but having multiple detailed comments may add noise over time without providing additional value. Since users typically rely on the main description when exploring the method, it might be more effective to consolidate this into a single, well-written summary comment.
```suggestion
/** Spec v4 (§5.1, §5.2): ValidationProcess (validation-request, GRANTOR/ECOSYSTEM) and DirectIssuance (issuance-request, HOLDER/ISSUER), ensuring type-safe flow handling. */
export enum VtFlowVariant {
  ValidationProcess = 'validation-process',
  DirectIssuance = 'direct-issuance',
}
```

packages/vt-flow/src/VtFlowModuleConfig.ts
```diff
@@ -0,0 +1,124 @@
...
+/** Fired on COMPLETED. Typical applicant use: link the credential to the DID Doc. */
+export type VtFlowOnCompletedHook = (ctx: VtFlowCredentialLifecycleContext) => Promise<void>
+
+export interface VtFlowBuildCredentialOfferContext {
+  agentContext: AgentContext
+  record: VtFlowRecord
+}
+
+export interface VtFlowCredentialOfferPayload {
+  // biome-ignore lint/suspicious/noExplicitAny: format depends on host agent registrations.
```
lotharking: We are not using Biome. If something needs to be ignored, please use ESLint ignore directives instead.

packages/vt-flow/src/VtFlowModule.ts
```diff
@@ -0,0 +1,270 @@
...
+   *  - a listener on `DidCommCredentialStateChanged` that maps
+   *    subprotocol transitions onto the vt-flow session via
+   *    `~thread.pthid` => `VtFlowRecord.threadId` (see
+   *    `doc/vt-flow-implementation.md`
+   *    §Superprotocol/Subprotocol Event Correlation);
+   *  - a listener on our own `VtFlowStateChanged` event so the
+   *    `onCompleted` callback fires exactly once per session.
+   */
+  public async initialize(agentContext: AgentContext): Promise<void> {
+    const messageHandlerRegistry = agentContext.dependencyManager.resolve(DidCommMessageHandlerRegistry)
```
lotharking, line 52: Add the API registration.
```suggestion
    dependencyManager.registerContextScoped(VtFlowApi)
    const messageHandlerRegistry = agentContext.dependencyManager.resolve(DidCommMessageHandlerRegistry)
```

lotharking:
Hey @tarunvaddeSoul, is this PR ready for review?

packages/vt-flow/src/VtFlowVariant.ts
tarunvadde: Good point. I will summarize these comments.

tarunvadde:
> Hey @tarunvaddeSoul, is this PR ready for review?

The implementation is still pending but your review would be helpful.

tarunvadde:
@lotharking, If possible, could you please review the `POST /v1/vt/onboard`API and let me know if I'm on the right track?

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingController.ts
```diff
@@ -0,0 +1,48 @@
...
+  public constructor(private readonly service: VtFlowOnboardingService) {}
+
+  @Post()
+  @ApiOperation({
+    summary: 'Start an Applicant-side vt-flow session against a remote Validator.',
+  })
+  @ApiBody({
+    schema: {
+      example: {
+        validatorDid: 'did:webvh:ecs-trust-registry.testnet.verana.network',
```
lotharking: IMHO, this should be `applicantDID` or something similar

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingController.ts
```diff
@@ -0,0 +1,48 @@
...
+        walletAgentPermId: 'wallet-agent-perm-42',
+        schemaBaseId: 'organization',
+        claims: {
+          name: 'Example Organization',
+          country: 'CH',
+        },
+      },
+    },
+  })
+  public async startOnboarding(@Body() body: VtFlowOnboardRequest): Promise<VtFlowOnboardResponse> {
```
lotharking: How do IR and VR differ?

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingService.ts
```diff
@@ -0,0 +1,199 @@
...
+          sessionUuid,
+          schemaId: body.schemaId as string,
+          agentPermId: body.agentPermId,
+          walletAgentPermId: body.walletAgentPermId,
+          claims,
+        })
+
+    // Tag `schemaBaseId` + `validatorDid` for the COMPLETED listener.
+    record.setTag('schemaBaseId', body.schemaBaseId)
+    if (body.validatorDid) record.setTag('validatorDid', body.validatorDid)
```
lotharking: If this onboarding is on the validator side, please check what happened when accepting the invitation. I remember it saves the DID.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingController.ts
tarunvadde: Did you mean the casing (Did vs DID), or are you asking for an applicantDid field? Currently validatorDid names which Validator to connect to, and the Applicant's DID is derived from the agent itself.

lotharking: I understant that you are creating the onboarding flow to handle this, but I don't think it's required at the moment.
According to spec 4.1 and 4.2, when a VS Agent starts, we need to set the `VS_AGENT_MODE` environment variable. Based on that, the agent should use `vtFlowApi` to call either `sendValidationRequest` or `sendIssuanceRequest`.

packages/vt-flow/src/VtFlowVariant.ts
tarunvadde: Done

packages/vt-flow/src/VtFlowModuleConfig.ts
tarunvadde: I saw biome being used here `packages/agent-sdk/src/did/CachedDocumentLoader.ts`. I replaced with `ESLint ignore` as per your suggestion.

packages/vt-flow/src/VtFlowModule.ts
tarunvadde: Done. Thanks.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingController.ts
tarunvadde: The presence of permId selects VR, schemaId selects IR.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingService.ts
tarunvadde: I'll drop the validatorDid tag in the next commit and have the listener resolve it via `agent.didcomm.connections.findById(record.connectionId).theirDid`. Thanks for the catch.

tarunvadde: Btw this is Applicant-side, not validator-side - this captures the remote Validator's DID.

packages/vt-flow/src/VtFlowVariant.ts
tarunvadde: Done in 63889d5 — applied the single-summary style to every file in the vt-flow package (VtFlowState, VtFlowApi, VtFlowService, messages, handlers, repository, errors). Thanks for the suggestion.

packages/vt-flow/src/VtFlowModuleConfig.ts
tarunvadde: Done in 63889d5 — replaced every `biome-ignore` in the vt-flow package with `eslint-disable-next-line`. Zero biome directives remain.

packages/vt-flow/src/VtFlowModule.ts
tarunvadde: Done in 63889d5 — added `dependencyManager.registerContextScoped(VtFlowApi)` in `register()`. That's also what the auto-offer event listener resolves.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowOnboardingController.ts
tarunvadde: Agreed. @genaris confirmed the same on our call. 

I've scrubbed all onboarding code from apps/vs-agent:

packages/agent-sdk/src/plugins/setupBaseDidComm.ts
```diff
@@ -57,6 +58,8 @@ export function setupBaseDidComm(options: BaseDidCommPluginOptions): BaseDidComm
               credentialFormats: [
                 new LegacyIndyDidCommCredentialFormatService(),
                 new AnonCredsDidCommCredentialFormatService(),
+                // Required for W3C JSON-LD credentials (vt-flow, ECS).
```
genaris:
```suggestion
```
No need to clarify

packages/agent-sdk/src/agent/VsAgent.ts
```diff
@@ -264,11 +269,15 @@ export class VsAgent<TModules extends BaseAgentModules = BaseAgentModules> exten
...
-    await didRepository.update(this.agentContext, record)
+    // Fresh did:web bootstrap: `record` is undefined here; the later
+    // `dids.create({...didDocument})` call persists the keys.
+    if (record) {
+      record.keys?.push({
+        kmsKeyId: key.keyId,
+        didDocumentRelativeKeyId: `#${publicKeyMultibase}`,
+      })
+      await didRepository.update(this.agentContext, record)
+    }
```
genaris: This is certainly a bug for did:web (and for did:webvh it just coincidentall work), but doesn't belong to this PR. Also, the comment is misleading, since dids.create() in WebDidRegistrar does not persist any key, so basically it prevents the crash but it does nothing in this case.

I raised [a PR to address this](https://github.com/verana-labs/vs-agent/pull/418) using a different approach. Hopefully it would work fine for all cases and we can skip this change here.

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -0,0 +1,349 @@
...
+  sessionUuid?: string
+  schemaId: string
+  agentPermId: string
+  walletAgentPermId: string
+  claims?: Record<string, unknown>
+}
+
+export interface OfferCredentialForSessionOptions {
+  vtFlowRecordId: string
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
genaris: Remove all these // eslint-disable statements unless you have a clear reason to justify it.

For example, in this case, credentialFormats can be clearly limited to the jsonld format

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -0,0 +1,349 @@
...
+    const record = await this.vtFlowService.getById(this.agentContext, vtFlowRecordId)
+    record.assertRole(VtFlowRole.Applicant)
+
+    if (!record.credentialExchangeRecordId) {
+      throw new CredoError(
+        `VtFlow record '${record.id}' has no linked credentialExchangeRecordId — did an offer arrive?`,
+      )
+    }
+
+    const credentialsApi = this.resolveCredentialsApi()
```
genaris: I don't see so much added value in `resolveCredentialsApi()` method.

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -0,0 +1,349 @@
...
+  public findByThreadId(threadId: string): Promise<VtFlowRecord | null> {
+    return this.vtFlowService.findByThreadId(this.agentContext, threadId)
+  }
+
+  public findAllByQuery(query: Query<VtFlowRecord>, queryOptions?: QueryOptions): Promise<VtFlowRecord[]> {
+    return this.vtFlowService.findAllByQuery(this.agentContext, query, queryOptions)
+  }
+
+  private resolveCredentialV2Protocol(): DidCommCredentialProtocol & {
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
genaris:
```suggestion
```

genaris: Same for all of these eslint-disable-next-line

packages/vt-flow/src/VtFlowConnectionState.ts
genaris, line 1: I'd remove everything related to `VtFlowConnectionState` from this module, leaving the whole DIDComm connection life cycle to the caller (see my main comment for more details).

genaris (changes requested):
I think in general the approach is OK, since it leaves every step open for the caller application so it can easily handle the protocol life cycle. 

I'm not convinced about managing/tracking DIDComm connection state within this module, since I feel it would be easier to handle it directly on the main application (VS Agent in this case). 

That being said, I think this module should check for VS-CONN-VS compatibility every time the protocol is started (either for Applicant and for Validator). For this I can imagine that we can replace the typical `connection.assertReady()` line by a convenient callback (such as `assertVerifiableService(connectionId)`) that can be defined as an API option or VtFlowModuleConfig parameter. This callback, implemented by the caller app, will be in charge to decide if the given DIDCommConnection meets the requirements of a VS. Probably at some point in the future we can integrate it directly on this module, but for the moment we can leave this implementation open.

tarunvadde:
> I think in general the approach is OK, since it leaves every step open for the caller application so it can easily handle the protocol life cycle.
> 
> I'm not convinced about managing/tracking DIDComm connection state within this module, since I feel it would be easier to handle it directly on the main application (VS Agent in this case).
> 
> That being said, I think this module should check for VS-CONN-VS compatibility every time the protocol is started (either for Applicant and for Validator). For this I can imagine that we can replace the typical `connection.assertReady()` line by a convenient callback (such as `assertVerifiableService(connectionId)`) that can be defined as an API option or VtFlowModuleConfig parameter. This callback, implemented by the caller app, will be in charge to decide if the given DIDCommConnection meets the requirements of a VS. Probably at some point in the future we can integrate it directly on this module, but for the moment we can leave this implementation open.

Thank you for your comments, @genaris. 

I have pushed the architectural changes from your review:
- Dropped VtFlowConnectionState. The module no longer tracks DIDComm connection state. 
- Added assertVerifiableService hook. Type lives on VtFlowModuleConfig. The module invokes it at the four session-starting sites the spec calls out: Applicant before sending validation-request or issuance-request, Validator on receipt of either. On rejection it throws a vt-flow.not-a-verifiable-service error. When the hook is undefined the module logs a warning and permits, so existing tests keep passing. Per your note that this is implemented by the caller app, the body is intentionally left out of this PR.

apps/vs-agent/tests/vtFlow.test.ts
```diff
@@ -0,0 +1,184 @@
...
+    expect(validatorRecord?.role).toBe(VtFlowRole.Validator)
+    expect(validatorRecord?.variant).toBe(VtFlowVariant.DirectIssuance)
+    expect(validatorRecord?.state).toBe(VtFlowState.Validating)
+    expect(validatorRecord?.threadId).toBe(applicantRecord.threadId)
+    expect(validatorRecord?.sessionUuid).toBe(applicantRecord.sessionUuid)
+    expect(validatorRecord?.schemaId).toBe(applicantRecord.schemaId)
+  })
+
+  it('§5.1 validation-request: Applicant VR_SENT, Validator auto-accepts + auto-marks-validated', async () => {
+    const validatedReached = waitForState(validator, VtFlowState.Validated)
```
lotharking: You should use `waitForEvent` on helpers

apps/vs-agent/package.json
```diff
@@ -86,6 +86,7 @@
     "yargs": "^17.3.1"
   },
   "optionalDependencies": {
+    "@verana-labs/credo-ts-didcomm-vt-flow": "workspace:*",
```
lotharking: Hey @genaris, is this optional?
I'm not sure about the name. Is it a protocol or a plugin?

packages/vt-flow/src/handlers/OobLinkHandler.ts
```diff
@@ -0,0 +1,16 @@
...
+
+/** Applicant-side inbound handler for `oob-link`; delegates to `VtFlowService.processReceiveOobLink`. */
+export class OobLinkHandler implements DidCommMessageHandler {
+  public supportedMessages = [OobLinkMessage]
+
+  public constructor(private readonly vtFlowService: VtFlowService) {}
+
+  public async handle(messageContext: DidCommMessageHandlerInboundMessage<OobLinkHandler>) {
+    await this.vtFlowService.processReceiveOobLink(messageContext)
+    return undefined
```
lotharking, line 14: I prefer to return `void`; returning undefined can cause confusion and errors
```suggestion
    return
```

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -0,0 +1,335 @@
...
+  DidCommMessageSender,
+  getOutboundDidCommMessageContext,
+} from '@credo-ts/didcomm'
+
+import { VtFlowModuleConfig } from './VtFlowModuleConfig'
+import { VtFlowRole } from './VtFlowRole'
+import { VtCredentialState } from './messages'
+import { VtFlowService } from './services'
+
+export interface SendValidationRequestOptions {
```
lotharking: Please remember to split responsibilities; I prefer this on a types file.

packages/vt-flow/src/VtFlowRole.ts
```diff
@@ -0,0 +1,5 @@
+/** Role a party plays in a vt-flow session; perspective-only, never on the wire. */
+export enum VtFlowRole {
```
lotharking: I prefer to group all these types in a single `types.ts` file.

packages/vt-flow/tests/VtFlowEnums.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    'COMPLETED',
+    'CRED_REVOKED',
+    'TERMINATED_BY_VALIDATOR',
+    'TERMINATED_BY_APPLICANT',
+    'ERROR',
+    'PERM_REVOKED',
+    'PERM_SLASHED',
+  ]
+
+  it('enumerates every Flow State from the spec', () => {
```
lotharking: I'm not sure this test is relevant. Could we add tests only when they provide value to the project?

packages/vt-flow/src/handlers/OobLinkHandler.ts
tarunvadde: Thanks for the suggestion. I'd lean towards keeping return undefined here, for two reasons:

1. Credo's own DIDComm handlers use it overwhelmingly.  Mirroring that convention keeps the package familiar to anyone reading both codebases.
1. The handle method's return type is Promise<DidCommOutboundMessageContext | undefined>, not Promise<void>. When the function explicitly may return undefined, returning the value undefined reads more directly than a bare return that the compiler coerces.

apps/vs-agent/tests/vtFlow.test.ts
tarunvadde: Done

apps/vs-agent/package.json
tarunvadde: @genaris @lotharking small clarification needed.

The optional dep is currently only needed by tests/vtFlow.test.ts. Should I move it to devDependencies since vt-flow isn't loaded at runtime in vs-agent today?

Per Ariel's earlier guidance to keep apps/vs-agent untouched I went with style scrubbing, but the existing optional-deps slot survived. Quick confirm and I'll fix in the same pass.

packages/vt-flow/tests/VtFlowEnums.test.ts
tarunvadde: Well, it doesn't serve any larger purpose but I just wanted to add it with the intent of covering some basic unit tests.

Removed it.

apps/vs-agent/package.json
lotharking: This should probably be included as part of the dependencies. The idea behind VS Agent is to always include the Verana setup, since it is required for handling credentials.
I think for now it's ok, but further we should move this on demand.

lotharking (approved):
If all is resolved, I think we could proceed with this and start testing on the VS Agent

apps/vs-agent/package.json
tarunvadde: Yes, will do that. Thanks, @lotharking.

lotharking:
Hey @tarunvadde, is this PR complete?

tarunvadde:
> Hey @tarunvadde, is this PR complete?

Yes.

genaris (approved):

PR #413, opened 2026-04-21
fix: remove mrtd auto-enable logic and add plugin load warnings

lotharking:
**Changes**
- Create validation for new vs agent images.
- According to this, the docs have been updated.


AirKyzzZ (approved):
LGTM

PR #415, opened 2026-04-22
fix: remove revocation registry on Delete credential types

lotharking:
**Changes**
- Remove revocation registry associate credential definition when this has been removed.

### DIFF
--- a/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeController.ts
+++ b/apps/vs-agent/src/controllers/admin/credentials/CredentialTypeController.ts
@@ -6,7 +6,6 @@ import {
   AnonCredsCredentialDefinitionRepository,
   AnonCredsKeyCorrectnessProofRecord,
   AnonCredsKeyCorrectnessProofRepository,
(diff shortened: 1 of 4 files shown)


genaris (commented):
I have no problem on doing this automatically when deleting a credential type, but maybe it would be more appropriate to do so only if a particular flag (e.g. "deleteAssociatedRevocationRegistries") is enabled. It is an edge case, but probably we would like to still be able to check status of credentials already issued (clients may have the credential type cached but not the revocation registry/status list).

BTW what happens in the case of Revocation Registry DELETE method? Is it properly deleting the associated attestedResource?

And what about the Revocation Status lists? Wouldn't be useful a way to specify that we want to clean them up as well? And maybe the tails files.

genaris:
@lotharking have you seen my comments regarding this PR?

lotharking:
Hey @genaris,
The DELETE method has already been implemented, and the service handles the deletion of all related resources.
For credential types, we’ve added a flag that allows controlling whether associated revocation registries are also deleted. This includes cleaning up related status lists and removing the corresponding tails files when those status lists are deleted.

genaris (approved):

PR #416, opened 2026-04-22
fix: update Credential Revocation swagger

### LINKED ISSUE #412: Wrong schema in API docs for /vt/revoke-credential
Shows { id: string }

It should be 

{
  "format": "anoncreds",
  "anoncredsRevocationRegistryDefinitionId": "did:webvh:xxx",
  "anoncredsRevocationRegistryIndex": 1
}

lotharking:


genaris (changes requested):
This has nothing to do with #416 . It is about the Revoke Credential DTO. Look at the Swagger API and you will see that the example is not what this POST actually expects.

We don't need to change the response from this method.

genaris (approved):
This will be a fix because docs will not generate a new version.

PR #422, opened 2026-04-29
feat: indexer notification handlers

lotharking:
**Changes**

* Extend §7 with a **customizable notification handler layer**
* Add handler interface + default implementation
* Enable overriding handlers per notification type
* Move from direct `applyActivity` to dispatch-based processing

> Based on spec (§7): each notification must map to a handler interface that can be overridden 
>
> Continuation of:
>
> * [https://github.com/verana-labs/vs-agent/pull/401](https://github.com/verana-labs/vs-agent/pull/401)
>
> The following were out of scope and deferred to:
>
> * §6 — Automatic VTJSC generation and DID Document publication https://github.com/verana-labs/vs-agent/pull/423
>
> This PR completes the handler architecture by introducing dispatch-based processing instead of state mirroring (`applyActivity`).


packages/agent-sdk/src/blockchain/handlers/defaultHandlers.ts
```diff
@@ -0,0 +1,156 @@
...
+import {
+  bumpActiveGfVersion,
+  upsertCredentialSchema,
+  upsertPermission,
+  upsertTrustRegistry,
+} from './stateMutations'
+
+/**
+ * Based on spec 7.2, these are the "default" handlers for indexer events emitted by the Verana blockchain.
+ */
```
genaris, line 11: Don't refer to "spec x.x" in the comments, since there is not a formal documentation for this feature. We made it based on an epic issue which is structured in a way that seems to be formal, but after this work is done we are not going to store the documentation exactly as it is written on the issue.

genaris (approved):
Looks promising!

tarunvadde (approved):

PR #423, opened 2026-04-30
feat: automatic vtjsc generation

### LINKED ISSUE #424: Credential Schema/VTJSC synchronization
When VPR integration is enabled, VS Agent needs to automatically create a Verifiable Trust JSON Schema credential upon the creation of a Credential Schema related to its DID on Verana network. For this purpose, it will need to query all CS and detect any new CS created using the indexer endpoint (as explained in section 6 of #402).

In this case, we need to think on the case where administrators manually manipulate linked JSCs by using `json-schema-credentials` endpoints. Would it make sense to keep POST/DELETE in this case? (i.e. could be possible for a given VS Agent to expose JSCs non related to Verana VPR?) If so, what if after deleting a JSC we want the agent to synchronize again with the VPR?

At first sight, it seems to me that it would be safer to just hide these endpoints when using this synchronization (i.e. when VERANA_* env variables are defined), and have an internal mechanism in VS Agent to make sure that all manually created JSC are deleted prior the initial sync. This would be the case of a migration from an existing VS Agent to a VPR-integrated VS Agent.

lotharking:
**Changes**

* Move `setupSelfTr.ts` to agent SDK
* Introduce `trustCredentialStore.ts` for credential persistence and handling
* Refactor `TrustService.ts` logic into `trustCredentialStore.ts` to share logic between VS-Agent and SDK
* Add `getChainId` to `VeranaChainService.ts`
* Update `stateMutations.ts` to support add/remove operations aligned with spec §6

> Based on spec (§6): automatic VTJSC generation and DID Document publication 
>
> Continuation of:
>
> * [https://github.com/verana-labs/vs-agent/pull/401](https://github.com/verana-labs/vs-agent/pull/401)
> [https://github.com/verana-labs/vs-agent/pull/422](https://github.com/verana-labs/vs-agent/pull/422)
>
> The following remains out of scope:
>


apps/vs-agent/src/controllers/admin/verifiable/IndexerDisabledGuard.ts
```diff
@@ -0,0 +1,15 @@
+import { CanActivate, ForbiddenException, Injectable } from '@nestjs/common'
+
+import { VERANA_INDEXER_BASE_URL } from '../../../config/constants'
+
+@Injectable()
+export class IndexerDisabledGuard implements CanActivate {
+  canActivate(): boolean {
+    if (VERANA_INDEXER_BASE_URL) {
+      throw new ForbiddenException(
+        'VTJSC lifecycle is managed automatically by the indexer. Manual create/delete is disabled.',
```
genaris: This comment is misleading. Anyone who read this thinks that the indexer is responsible for creating and deleting JSC credentials.

You can say "managed automatically from Verana VPR events" or similar, but not the indexer by itself.

apps/vs-agent/src/controllers/admin/verifiable/TrustController.ts
```diff
@@ -315,7 +316,8 @@ export class TrustController {
       'Deletes a stored Verifiable Trust Json Schema Credential (VTJSC) associated with the specified schema identifier (`schemaId`). ' +
       'A JSON Schema Credential defines the structure and validation rules for a Verifiable Trust Credential (VTC). ' +
       'Removing a JSC also invalidates any Verifiable Trust Credentials that rely on it. ' +
-      'This operation follows the [Verifiable Trust Specification](https://verana-labs.github.io/verifiable-trust-spec/#json-schema-credentials).',
+      'This operation follows the [Verifiable Trust Specification](https://verana-labs.github.io/verifiable-trust-spec/#json-schema-credentials).\n\n' +
+      '**Disabled when `VERANA_INDEXER_BASE_URL` is configured.** VTJSC lifecycle is managed automatically by the indexer.',
```
genaris: Same as above.

apps/vs-agent/src/controllers/admin/verifiable/TrustController.ts
```diff
@@ -359,6 +362,7 @@ export class TrustController {
   The **issuer DID** of the VTJSC MUST be the **same DID** as the Ecosystem DID of the Trust Registry that created the referenced CredentialSchema in the ledger.
 
   VTJSCs issued by any other DID will be be considered invalid by trust resolvers.
+  **Disabled when \`VERANA_INDEXER_BASE_URL\` is configured.** VTJSC lifecycle is managed automatically by the indexer.
```
genaris: Same as above

packages/agent-sdk/src/utils/trustCredentialStore.ts
```diff
@@ -0,0 +1,307 @@
...
+  createPresentation,
+  getVerificationMethodId,
+  mapToSelfTr,
+  presentations,
+  signerW3c,
+} from './setupSelfTr'
+
+export interface TrustMetadataEntry {
+  credential: any
+  verifiablePresentation: any
```
genaris: In our basic coding rules we don't allow `any` and `force non-null`. In this PR I noticed that those are not being applied. Can you check what is going on?

genaris (commented):
Maybe most of the work on this PR is a refactor, but to me it is clearly a feature, since here we are introducing a new behaviour.

packages/agent-sdk/src/utils/trustCredentialStore.ts
lotharking: Sorry, this was a legacy test

genaris (approved):

Issue #425, opened 2026-04-30
Administration API (§8 of v4 VPR Integration Spec)


tarunvadde:
## Context

Section 8 of the v4 VPR Integration Spec (#402) requires VS Agent to expose a secure Administration API that lets authorized members of the agent's Corporation remotely query and manage agent state.

### Phase 0: #426 (Validation Process Flow basic)

#426 implements the §5.1 validation flow end-to-end (chain integration via `start-perm-vp` / `set-perm-vp-validated`, DIDComm vt-flow protocol, and the manual lifecycle Admin API endpoints needed to drive the flow). Those manual lifecycle endpoints are the natural home for the §5.1 write actions originally scoped to this issue. Phase 0 lands first; once it does, the controller scaffolding and the §5.1 lifecycle endpoints exist on the agent and the remaining §8 surface here is small.

### Phase 1: Queries (read-only)

Endpoints under `/v1/vt-flows`:

- [ ] `GET /v1/vt-flows`: list flows, filterable by `flowState`, `peerDid`, `permId`, `schemaId`, `variant`, `role`
- [ ] `GET /v1/vt-flows/:vtFlowRecordId`: flow details including claims, proofs, OOB URLs
- [ ] `VtFlowNestPlugin` registering the controller, gated on `ENABLED_PLUGINS.includes('vt-flow')`

Depending on how #426 lands, some of this scaffolding may already exist as a side effect there; this phase fills any gaps.

**Out of scope for Phase 1:** `connectionState` filter (defer until we expose connection state through the spec lens), permission session details in the response (require chain queries; separate concern).

### Phase 2: Actions for the rest of the §8 surface

Endpoints under `/v1/vt-flows`, covering everything not addressed by #426:

- [ ] §5.2 Direct Issuance lifecycle (Validator: accept issuance request, offer credential; Applicant: accept credential)
- [ ] Validator: edit claims, send OOB link, revoke credential, terminate flow
- [ ] Applicant: terminate flow
- [ ] Audit metadata persisted on each action (caller's Verana account once Phase 3 lands)
- [ ] `connectionState` filter from Phase 1 backlog

### Phase 3: Auth layer (deferred)

Per @genaris, this depends on features that aren't fully stable yet, including the recent `Authority` to `Corporation` refactor. Tracking as a placeholder for now; will be revisited once the underlying features stabilize:

- [ ] ADR-036 signature challenge for Verana-account authentication
- [ ] Corporation membership check on the calling Verana account
[...]

Issue #402, opened 2026-04-17
Verana VPR Integration


genaris:
## 1. Overview

Resources created in the Verana ledger (a VPR) are linked to DIDs that represent VS-Agents. For this reason, a VS-Agent MUST receive notifications of changes in the ledger that are directly or indirectly linked to its DID, and update its state accordingly.

**Examples:**

- **Trust registry schema addition** — A new credential schema is created for a trust registry. The VS-Agent whose DID owns that trust registry is notified and automatically creates the corresponding VTJSC, publishing it in its DID Document.
- **Validation process lifecycle** — An applicant initiates a validation process to obtain a HOLDER permission from an ISSUER for a given credential schema. The applicant creates the Validation Process on the Verana ledger. The VS-Agents of both applicant and validator (ISSUER) are notified and begin a userland validation flow over DIDComm. As the on-chain permission state changes, the respective VS-Agents receive further notifications and execute follow-up tasks (e.g., continuing the DIDComm exchange, issuing the credential).

```mermaid
flowchart LR
    VPR["Verana Ledger<br/>(VPR)"]
    IDX["Indexer<br/>(WebSocket)"]
    VS["VS-Agent<br/>(TR Owner)"]
    DID["DID Document"]

    VPR -- "new credential schema<br/>created in TR" --> IDX
    IDX -- "notification" --> VS
    VS -- "generate VTJSC +<br/>publish VP" --> DID
```

*Figure 1a — Trust registry schema addition. A new credential schema is created on-chain; the Indexer notifies the owning VS-Agent, which generates the corresponding VTJSC and publishes it in its DID Document.*

```mermaid
flowchart LR
    VPR["Verana Ledger<br/>(VPR)"]
    IDX["Indexer<br/>(WebSocket)"]
    VSA["VS-Agent A<br/>(Applicant)"]
    VSB["VS-Agent B<br/>(Validator / ISSUER)"]

    VPR -- "permission event" --> IDX
    IDX -- "perm change notifications" --> VSA
    IDX -- "perm change notifications" --> VSB
    VSA <-- "DIDComm<br/>(validation + issuance)" --> VSB
```

*Figure 1b — Validation process lifecycle. The applicant creates a Validation Process on-chain; both VS-Agents are notified and coordinate over DIDComm. As the on-chain permission state changes, further notifications trigger follow-up tasks.*

[...]

PR #430, opened 2026-05-01
feat(vt-flow): onboarding-process flow + v4 participant/ecosystem migration

### LINKED ISSUE #426: Validation Process Flow implementation (basic)
The goal is to do an initial implementation of the validation process flow, based on the details written in #402 (section 5.1)

## Applicant side

The flow will start at the **Applicant VS Agent**, by calling an Admin API method (following the ideas from #425, it can be a POST /v1/vt-flows), which receives as inputs some basic info:
- Validator Perm ID
- Claims for the credential: initially we are focused on Org or Persona ECS schemas, but defined like this, it can be actually used for any schema

Then, VS Agent will internally query indexer to check that the specified Perm ID is active, and retrieve validator DID from its data.

It will execute `start-perm-vp` to intiate a validation process to get a permission of type "HOLDER" for its own DID. This will return VS Agent's holder permission ID.

Now it will be ready to start the vt-flow DIDComm protocol, sending the message to validator DID and passing to the `validation-request` message the holder permission ID, a random session UUID and the claims for the credential (proofs, agent_perm_id and wallet_agent_perm_id are not used).

It will keep idle until validator processes the request and creates the credential offer (o rejects the request). Based on the DIDComm message events received from vt-flow protocol, it will perform the corresponding checks (such as verify validator + digest) and update vt-flow record state.

Admin API POST methods must be added to manually continue the flow (e.g. accept credential), after it should update vt-flow record state and send DIDComm messages/update Linked VPs on its DID Document as needed.

## Validator side

Validator agent will receive the `validation_request`, and it will remain idle until a POST Admin API method (to be defined) sets it as validated (executes set-perm-vp-validated + calls corresponding vt-flow DIDComm module API method). At this point it will generate the credential (following the sign + compute digest procedure), execute createOrUpdate PermissionSession and finally send the Credential Offer through DIDComm.

It will handle further DIDComm events to determine when the credential has been accepted and therefore update the vt-flow record internal state.

## Notes

- This is the initial, minimum implementation where we expect that all "optional" steps of the sequence diagram are not needed. But in the implementation it would be nice to consider that they might be added later.
- At the moment, neither of the agents is going to perform the VS-CONN-VS checks to resolve trust. This will be added as a further issue
- Since we are focusing on the basic implementation, we can implement this with the current VS Agent status as is. So we can do it using DIDComm V1 and implicit invitations.

tarunvadde:
Basic onboarding-process flow (the old "validation process", renamed in [#457](https://github.com/verana-labs/vs-agent/issues/457)) in the SDK on the v4 chain, plus the admin endpoint to drive it. Covers #426 and the first slice of [#425](https://github.com/verana-labs/vs-agent/issues/425)'s admin API. It also carries the SDK-wide v4 rename, the flow can't run on the old v3 chain, so the migration lands here.

In this PR:

- Validate endpoint POST /v1/vt/flows/{participant_session_id}/validate: signs the credential, marks the participant validated on-chain, offers it over Issue Credential V2. (No start/accept endpoint, start is indexer-driven, accept is automatic.)
- VtFlowOrchestrator: start -> onboarding-request -> validate -> offer -> accept -> publish as LinkedVP.
- V4 chain migration on verana-types dev.16 (pp/ec messages, no version mixing).
- SDK-wide rename: permission -> participant, trust registry -> ecosystem across the chain service, indexer, handlers, orchestrator, and the vt-flow admin.
- Indexer on real v4: REST /verana/pp/v1/ + /verana/ec/v1/, v4 wire fields (op_state, ecosystem_id, validator_participant_id), and the v4/indexer/subscribe WS protocol.
- On-chain ParticipantSession created with the credential digest, which is computed over the credential content with the proof excluded so re-signing at issuance doesn't change it.

Verified end to end against a live v4 node + indexer (testcontainers): every chain field, indexer REST field/path, and indexer event entity_type + action matches. One temporary bit in the comment below.


packages/agent-sdk/src/blockchain/VeranaChainService.ts
```diff
@@ -74,12 +75,13 @@ export class VeranaChainService {
     this.permQuery = new PermQueryClientImpl(rpc) as PermQueryClient
   }
 
-  // Query API (No signed)
-  async getPermission(id: Long): Promise<unknown> {
-    return this.permQuery.GetPermission({ id })
+  // Query API (unsigned)
```
lotharking, line 119: Please be carefull with this change. It doesn't introduce new behaivor, which may add unnecessary noise and make the review more complex

apps/vs-agent/src/controllers/admin/vt-flow/dto/validate-flow.dto.ts
```diff
@@ -0,0 +1,25 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { IsArray, IsNotEmpty, IsOptional, IsString } from 'class-validator'
+
+export class ValidateFlowDto {
+  @ApiProperty({
+    description:
+      'Identifier of the JsonSchemaCredential the issued credential should reference (per VT-CRED-W3C).',
+    example: 'https://example.test/schemas/organization-credential.json',
```
lotharking:
```suggestion
    example: 'https://localhost:3001/vt/schemas-organization-jsc.json',
```

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+export class VtFlowsService {
+  public constructor(@Inject(VsAgentService) private readonly agentService: VsAgentService) {}
+
+  public async startValidationProcess(input: StartValidationProcessDto): Promise<VtFlowRecordDto> {
+    const agent = await this.agentService.getAgent()
+    const chain = this.requireChain(agent)
+    if (!agent.did) throw new BadRequestException('Agent has no public DID')
+
+    const validatorPermId = this.parsePermId(input.validatorPermId, 'validatorPermId')
+    const validatorPerm = await chain.getPermission(validatorPermId)
```
lotharking: I'd prefer using [VeranaIndexerService.ts](https://github.com/verana-labs/vs-agent/pull/423/changes#diff-d7c8bed487f1bbe8aa7cbda990e98cf0e8e71ee230813740a7a8fb8fccc4af3e) to retrieve permissions. If we query the ledger here instead, we risk inconsistencies with the indexer state

lotharking (changes requested):

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsController.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  @Post('/')
+  @HttpCode(HttpStatus.CREATED)
+  @ApiOperation({
+    summary: 'Start a §5.1 Validation Process',
+    description:
+      'Applicant action. Runs `start-perm-vp` on-chain, opens an implicit DIDComm connection to the validator, and dispatches the vt-flow `validation-request`.',
+  })
+  @ApiCreatedResponse({ type: VtFlowRecordDto })
+  public startValidationProcess(@Body() body: StartValidationProcessDto): Promise<VtFlowRecordDto> {
```
lotharking: Hey @genaris, I'm not completely sure about it. 
When we start the validation process, does the agent require us to do this? My question comes from the frontend perspective (and also the ledger), where it's common to initiate the process from there; the flow usually starts on that side. On the other hand, the agent seems to just listen to the indexer WebSocket and then begins sending the `DIDComm connect.`

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
tarunvadde: Since #423 is not merged yet, I added a temporary chain query but will switch to `VeranaIndexerService.getPermission` once [#423](https://github.com/verana-labs/vs-agent/issues/423) lands. 

Any other suggestions than this?

packages/agent-sdk/src/blockchain/types.ts
```diff
@@ -1,5 +1,33 @@
 import { BaseLogger, JsonObject } from '@credo-ts/core'
 
+/** Permission type values per `verana.perm.v1.PermissionType`. */
+export const PermissionType = {
+  UNSPECIFIED: 0,
```
lotharking: I'd prefer not mapping this state, this throw and error if it isn't a valid state
```suggestion
```

packages/agent-sdk/src/blockchain/VeranaChainService.ts
tarunvadde: The typed returns (`Permission | undefined, Permission[]) `are needed for `VtFlowsService` to read `permission.did, permission.revoked, permission.id `etc. without `as Permission` casts at every call site. Happy to revert and put the casts in the consumer if you'd prefer.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+    const subjectDid = holderPerm.did
+    const didRecords = await agent.dids.getCreatedDids({ did: agent.did })
+    const didRecord = didRecords[0]
+    if (!didRecord) throw new BadRequestException('Agent DID record not found')
+    const verificationMethodId = getVerificationMethodId(agent.config.logger, didRecord)
+
+    const claims = (record.claims ?? {}) as JsonObject
+    const unsignedCredential = createCredential({
+      id: `${agent.did}#${utils.uuid()}`,
+      type: input.credentialType ?? ['VerifiableCredential', 'VerifiableTrustCredential'],
```
lotharking: I'm not sure; this could be error-prone. Let's leave it as an input attribute since it's part of the validations; for now, I'd prefer the default one

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+
+    const signed = await signerW3c(
+      agent,
+      JsonTransformer.fromJSON(unsignedCredential, W3cCredential),
+      verificationMethodId,
+    )
+    const digest = generateDigestSRI(JSON.stringify(signed.jsonCredential))
+
+    await chain.setPermissionVPToValidated({
```
lotharking: What about error control or a logger? This step could fail, and we need to control this.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+    }
+    return n
+  }
+
+  private async findHolderPermission(
+    chain: VeranaChainService,
+    did: string,
+    validatorPermId: number,
+  ): Promise<Permission | undefined> {
+    const perms = await chain.findPermissionsWithDID({ did })
```
lotharking: This validation againt the indexer to avoid sync problems

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+    const ready = await agent.didcomm.connections.returnWhenIsConnected(connectionRecord.id)
+
+    const vtFlowApi = this.resolveVtFlowApi(agent)
+    const record = await vtFlowApi.sendValidationRequest({
+      connectionId: ready.id,
+      sessionUuid: input.sessionUuid ?? utils.uuid(),
+      permId: String(holderPerm.id),
+      agentPermId: '0',
+      walletAgentPermId: '0',
+      claims: input.claims,
```
lotharking: Claims need to be validated against the schema again before sending them to vtFlow

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -0,0 +1,237 @@
...
+      id: input.credentialSchemaCredentialId,
+      type: 'JsonSchemaCredential',
+    }
+
+    const signed = await signerW3c(
+      agent,
+      JsonTransformer.fromJSON(unsignedCredential, W3cCredential),
+      verificationMethodId,
+    )
+    const digest = generateDigestSRI(JSON.stringify(signed.jsonCredential))
```
lotharking: Remember, this needs to be checked on the validator side

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
lotharking, line 1: Remember, once the credential has been valid, you must include the credential in the DID Document

lotharking (changes requested):

packages/agent-sdk/src/blockchain/VeranaChainService.ts
lotharking: Got it

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsController.ts
tarunvadde: Good question. The current PR follows the explicit ask in #426:

> "The flow will start at the Applicant VS Agent, by calling an Admin API
> method (following the ideas from #425, it can be a POST /v1/vt-flows)"

So the agent owning the flow start is what the issue spec'd. The
frontend-driven alternative you mention is also a valid pattern (frontend
broadcasts `start-perm-vp` itself, agent reacts to the indexer event), but
it's a scope shift. I'll let @genaris confirm which model is preferred.
Happy to rework if Ariel wants the indexer-event-driven path instead.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
tarunvadde: Reading §5.1 step 3 in #402, claim validation is on the validator side:

> "The applicant MAY also include credential claims [...] The validator
> MUST either accept the information and proceed, or refuse it with an
> error code and descriptive error message."

So the validator is the one that validates claims and rejects bad ones.
For Phase 0 we're keeping that responsibility on the validator.

Adding applicant-side pre-validation as a nice-to-have is doable (fetch the
schema from the indexer, run JSON-schema validation with `ajv`), but it's
extra work outside the minimum scope of #426. Happy to add it as a follow-up
issue if you'd prefer that flow.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
tarunvadde: Could you clarify what you mean? Line 142 IS the validator-side digest
computation per §5.1 step 6 ("The validator generates and signs the
credential, and computes the digest"). The applicant-side check happens
later at step 9 when the applicant verifies the digest matches what's on
the chain.

Did you mean the validator should also re-verify the digest before
broadcasting `set-perm-vp-validated`, or something else? Want to make sure
I'm reading this right.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -91,6 +94,117 @@ export function upsertPermission(
...
+    permId,
+    async (record, service, agentContext) => {
+      await service.updateState(agentContext, record, VtFlowState.TerminatedByApplicant)
+      return 'TERMINATED_BY_APPLICANT'
+    },
+    'Failed to terminate record',
+  )
+}
+
+export async function startPermissionVPAutoFlow(agent: VsAgent, activity: IndexerActivity): Promise<void> {
```
```suggestion
export async function startParticipantOPAutoFlow(agent: VsAgent, activity: IndexerActivity): Promise<void> {
```

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -91,6 +94,117 @@ export function upsertPermission(
...
+  const chain = agent.veranaChain
+  if (!chain) return
+  const applicantPermId = Number(activity.entity_id)
+  if (!Number.isFinite(applicantPermId)) return
+  const holderPerm = await chain.getPermission(applicantPermId)
+  if (!holderPerm || holderPerm.did !== agent.did) return
+  try {
+    const orchestrator = new VtFlowOrchestrator(agent)
+    await orchestrator.startValidationProcess({ applicantPermId })
```
```suggestion
    await orchestrator.startOnboardingProcess({ applicantPermId })
```

lotharking (commented):
Please review the nomenclature according to the specification you are using. If it is v4, replace "permission" with "participant" and "trust registry" with "ecosystem".

apps/vs-agent/tests/helpers/events.ts
genaris, line 1: Do we really need to have these testing helpers and classes under `src/utils`? Why don't we just put them under `tests` (our usual practice in other repos), so we can easily exclude them from the bundled npm package?

.github/workflows/ci.yml
```diff
@@ -14,4 +14,4 @@ concurrency:
 
 jobs:
   ci:
-    uses: 2060-io/organization/.github/workflows/2060-linter-call.yml@main
+    uses: 2060-io/organization/.github/workflows/2060-linter-call.yml@main
```
genaris: Why this line removal? Was done automatically by biome or something?

apps/vs-agent/src/config/constants.ts
```diff
@@ -54,14 +53,6 @@ export const askarPostgresConfig: AskarPostgresStorageConfig = {
   },
 }
 
-export const keyDerivationMethodMap: {
```
genaris, line 57: Actually, I think this map shoul belong here instead of agent-sdk, as it was created to process vs-agent environment variables. So in my opinion, we should leave the definition here and remove from agent-sdk package.

packages/agent-sdk/tests/e2e/VeranaTestChain.ts
lotharking, line 1: Is this file required? I think these methods are already being used in ´VeranaChainService´. Please review whether we can remove this file and use the service instead, in order to reduce duplicated logic.

tarunvadde: Hey Andres, the two classes actually don't overlap, so there's nothing duplicated to remove. 

None of the VeranaTestChain methods exist in the service. The e2e uses both on purpose: VeranaTestChain sets up the world, VeranaChainService runs the flow against it.

The only real overlap is around 10 lines of cosmjs client boilerplate, and even that diverges (the test registers the group module, the service doesn't), so sharing it isn't worth it I believe.

Please let me know if I'm missing something or else do you want me to drop the "move this into the SDK" TODO in the e2e README?

apps/vs-agent/src/controllers/admin/vt-flow/dto/vt-flow-record.dto.ts
```diff
@@ -0,0 +1,22 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { VtFlowRole, VtFlowVariant, type VtFlowState } from '@verana-labs/credo-ts-didcomm-vt-flow'
+
+export class VtFlowRecordDto {
+  @ApiProperty() vtFlowRecordId!: string
```
genaris:
```suggestion
  @ApiProperty() id!: string
```

I think it is more consistent to respect the original fields of the record (e.g. createdAt, updatedAt, state).

apps/vs-agent/src/main.ts
```diff
@@ -180,6 +180,7 @@ const run = async () => {
     ...(ENABLED_PLUGINS.includes('messaging') ? [MessagingPlugin] : []),
     ...(chatModule ? [chatModule.ChatPlugin] : []),
     ...(mrtdModule ? [mrtdModule.MrtdPlugin({ masterListCscaLocation: MASTER_LIST_CSCA_LOCATION })] : []),
+    ...(ENABLED_PLUGINS.includes('vt-flow') ? [VtFlowNestPlugin] : []),
```
genaris: Since vt-flow is going now to be part of the core functionalities of VS Agent, I think it is appropriate to include it by default from now on.

apps/vs-agent/tests/__mocks__/global-setup.ts
```diff
@@ -1,3 +1,4 @@
+import '@openwallet-foundation/askar-nodejs'
```
genaris, line 1: Is this still needed? I thought it was fixed in one of the recent askar versions (which is used by credo 0.7.0)

packages/agent-sdk/README.md
```diff
@@ -95,23 +95,23 @@ import type { IndexerActivity, IndexerDispatchContext } from '@verana-labs/vs-ag
...
-  await agent.config.logger.info(`[Block ${context.blockHeight}] Permission ${status}: ${id}`)
+  await agent.config.logger.info(`[Block ${context.blockHeight}] Participant ${status}: ${id}`)
   
   // Emit to your service
-  await sendWebhookEvent('https://your-service.com/perms', {
-    type: 'permission_status_changed',
-    permissionId: id,
+  await sendWebhookEvent('https://your-service.com/participants', {
+    type: 'participant_status_changed',
+    participantId: id,
```
genaris: This is not part of the SDK anymore so I think it would be better to just clean up this README and remove any reference tosendWebhookEvent.

genaris (approved):
Some suggestions, not blocking but nice to handle before merging. We can address the FIXMEs later.

apps/vs-agent/tests/helpers/events.ts
tarunvadde: Done

.github/workflows/ci.yml
tarunvadde: The trailing newline at EOF got dropped accidentally. Thanks for pointing.

apps/vs-agent/tests/__mocks__/global-setup.ts
tarunvadde: Checked, still needed. Dropping it fails all 11 vs-agent unit tests with `Cannot read properties of undefined (reading 'keyGetJwkSecret') `in `AskarKeyManagementService`, the native binding isn't registered without it. 0.7.0 didn't change that here, so keeping it.

This seems to be related to https://github.com/openwallet-foundation/credo-ts/issues/2597. Please let me know if I'm missing something.

genaris: I thought it was related to this https://github.com/openwallet-foundation/askar-wrapper-javascript/issues/77, which was fixed in Askar wrapper v0.6.0. I don't think it is a blocker for us but curious to see why this wasn't solved yet.

genaris (approved):

apps/vs-agent/tests/__mocks__/global-setup.ts
tarunvadde: I am still not sure but I think I found why https://github.com/openwallet-foundation/askar-wrapper-javascript/issues/77 being fixed in 0.6.0 doesn't actually reach us: the fix is the new `NativeAskar.instance` getter, but `@credo-ts/askar` never adopted it. `AskarKeyManagementService` (and `kms/crypto/deriveKey`) still import the deprecated `askar` binding (`export let askar`, reassigned by `register()` after module eval), so credo's ESM KMS snapshots it as `undefined` before askar-nodejs's side-effect import runs -> `Cannot read properties of undefined (reading 'keyGetJwkSecret')`. `NativeAskar` appears nowhere in the package.

I did try patching this as a workaround in vs-agent and all the tests pass without the import at the top of the file.

genaris: Ok that's a good point. If it works fine please file a PR on Credo. I think it also worth to check if the current patch we are applying for webvh module (patches/@credo-ts__webvh@0.7.1-pr-2704-20260610194233.patch) is still needed and, if so, if it should also go upstream.

tarunvadde: Sure. Will take it upstream.

PR #431, opened 2026-05-01
feat(agent): add DIDComm v2 support

### LINKED ISSUE #427: Add DIDComm V2 support
We'll need to upgrade Credo to the latest alpha of PR 2704. We can leverage the existing work on didcomm-mediator to support both DIDComm V1 and DIDComm V2 connections in VS Agent's DID Document(support of each can be toggled by environment variables), but ultimately we are interested in supporting DIDComm V2 only.

We'll need to revisit all endpoints related to invitations to verify that it is possible to select the DIDComm version we want to use.

This is marked as part of VPR integration project because, in practice, we want to make Agent to Agent communication exclusively on DIDComm V2 (although it is still technically possible to keep using V1).

tarunvadde:
Closes #427.

## Summary
- Bumps Credo to `0.7.1-pr-2704-20260507141408`.
- New env var: `AGENT_DIDCOMM_VERSIONS` (default `v1,v2`)
- DID Document emits `DidCommV1Service` and/or `NewDidCommV2Service` per the flags.
- Invitation endpoints accept optional `didCommVersion`. 

## Validation
- Tested all the scenarios mentioned by @genaris 

> - An existing VS Agent can upgrade to this safely, and if I don't set any new environment variable it will add the DIDComm V2 services to the DID Document. It should check its DID Document services and check that they match the ones set up on AGENT_DIDCOMM_VERSIONS. If they don't, update the DID Document accordingly. We did a similar logic for did:web to did:webvh migration so maybe something done there can be reused/refactored
> - Hologram 2.13 (DIDComm V1 through production mediator) can connect and interact with it seamlessly (you can use /examples/chatbot docker compose for easy testing of basic messaging an credential issuance/proof request)
> - Hologram dev (DIDComm V2 through dev mediator) can also connect and interact with it. Since it also supports V1 and V2 by default, it should detect the V2 services and connect to it on V2 directly
> - You can create an invitation/presentation-request in both V1 and V2 and Hologram dev can properly scan both of them
> - You can create an invitation/credential-offer in both V1 and V2 and Hologram dev can properly scan both of them


packages/agent-sdk/src/utils/agent.ts
```diff
@@ -16,23 +16,38 @@ export async function createInvitation(options: {
...
 }) {
-  const { agent, messages, useLegacyDid, invitationBaseUrl, imageUrl } = options
+  const { agent, messages, useLegacyDid, invitationBaseUrl, imageUrl, didCommVersion } = options
+
+  if (didCommVersion === 'v2' && messages && messages.length > 0) {
+    throw new Error(
+      'DIDComm v2 out-of-band invitations do not support attached messages. ' +
+        'Use v1 for credential-offer / presentation-request flows, or send the message after connection forms.',
+    )
+  }
```
genaris: Can you elaborate about this? From [the DIDComm spec](https://identity.foundation/didcomm-messaging/spec/#invitation) I don't see any limitation to this. Is that we miss support for that in Credo?

In any case, we will need to think on a way to let users create their QR codes using `invitation/presentation-request` or `invitation/credential-offer`, in such a way that apps like Hologram will be able to show the request. If that represents that two or more DIDComm messages must be exchanged under the hood it is fine, as long as user experience is not significantly changed.

apps/vs-agent/src/config/constants.ts
```diff
@@ -76,6 +76,10 @@ export const ADMIN_LOG_LEVEL = process.env.ADMIN_LOG_LEVEL
 export const USE_CORS = Boolean(process.env.USE_CORS || false)
 export const ENABLE_PUBLIC_API_SWAGGER = !(process.env.ENABLE_PUBLIC_API_SWAGGER === 'false')
 
+// DIDComm versions
+export const DIDCOMM_V1_SUPPORT = process.env.DIDCOMM_V1_SUPPORT !== 'false'
+export const DIDCOMM_V2_SUPPORT = process.env.DIDCOMM_V2_SUPPORT === 'true'
```
genaris: For consistency I think these flags must be prefixed with AGENT, e.g. AGENT_DIDCOMM_V1_SUPPORT and AGENT_DIDCOMM_V2_SUPPORT. And maybe we can do like we did in Hologram, where there is a single variable with a list of supported versions. For example AGENT_DIDCOMM_VERSIONS=v1,v2.

In any case these settings are part of the advanced ones (don't forget to reference them on the README).

I think that we can enable both by default, since here we are not worried so much about backwards compatibility (at least not in default behaviour).

genaris (changes requested):

packages/agent-sdk/src/utils/agent.ts
tarunvadde: You're right, this is a Credo PR 2704 gap, not a spec limitation. `DidCommOutOfBandInvitationV2` has no `attachments` field and `createV2Invitation` silently drops `config.messages`. I'll add v2 OOB attachment support upstream in PR 2704 separately. Once that ships and we bump the alpha here, this rejection goes away and the credential-offer / presentation-request endpoints produce single-QR v2 invitations matching the v1 UX.

genaris: Great, let's do the updates on Credo then and we can upgrade it on this branch once it is done.

packages/agent-sdk/src/agent/VsAgent.ts
```diff
@@ -259,17 +263,38 @@ export class VsAgent<TModules extends BaseAgentModules = BaseAgentModules> exten
...
-        routingKeys: [], // TODO: Support mediation
-        recipientKeys: [keyAgreementId],
-        accept: ['didcomm/aip2;env=rfc19'],
-      })
+    const didcommVersions = this.didcomm!.config.didcommVersions
+    const includeV1 = didcommVersions.includes('v1')
+    const includeV2 = didcommVersions.includes('v2')
+    const services: (DidCommV1Service | NewDidCommV2Service)[] = []
+
+    this.didcomm!.config.endpoints.forEach((endpoint, index) => {
```
lotharking, line 287: Is it possible for the same agent to change from v1 to v2? Is this controlled here?

tarunvadde: Yes. `getDidCommServices` returns the desired V1 and/or V2 set from `AGENT_DIDCOMM_VERSIONS`. On startup, `VsAgent.initialize` compares the published DID Document's DIDComm services against this set, and if they differ it filters out the managed types (`MANAGED_DIDCOMM_SERVICE_TYPES` covers both `DidCommV1Service` and `NewDidCommV2Service`) and replaces them via `dids.update`. So flipping `AGENT_DIDCOMM_VERSIONS` and restarting updates the published doc. Keys aren't rotated since Ed25519 and X25519 are created together at first setup.

packages/agent-sdk/src/plugins/setupBaseDidComm.ts
```diff
@@ -32,6 +34,12 @@ export interface BaseDidCommPluginOptions {
   walletConfig: AskarModuleConfigStoreOptions
   publicApiBaseUrl: string
   endpoints: string[]
+  /**
+   * DIDComm envelope versions the agent supports. Defaults to `['v1']`.
+   * Including `'v2'` makes the agent accept inbound V2 envelopes and prefer V2 when sending
+   * to a connection that supports it.
+   */
```
genaris: Please avoid excess of comments. Let's maintain the verbosity at minimum.

apps/vs-agent/src/config/constants.ts
```diff
@@ -76,6 +76,13 @@ export const ADMIN_LOG_LEVEL = process.env.ADMIN_LOG_LEVEL
 export const USE_CORS = Boolean(process.env.USE_CORS || false)
 export const ENABLE_PUBLIC_API_SWAGGER = !(process.env.ENABLE_PUBLIC_API_SWAGGER === 'false')
 
+// DIDComm envelope versions accepted and sent by the agent. Comma-separated list of 'v1' / 'v2'.
+// Defaults to 'v1,v2'.
```
genaris: It is enough to mention how it works straight on the README

genaris (changes requested):
Code looks generally OK to me, but before merging this I want to make sure that:

- An existing VS Agent can upgrade to this safely, and if I don't set any new environment variable it will add the DIDComm V2 services to the DID Document. It should check its DID Document services and check that they match the ones set up on AGENT_DIDCOMM_VERSIONS. If they don't, update the DID Document accordingly. We did a similar logic for did:web to did:webvh migration so maybe something done there can be reused/refactored
- Hologram 2.13 (DIDComm V1 through production mediator) can connect and interact with it seamlessly (you can use /examples/chatbot docker compose for easy testing of basic messaging an credential issuance/proof request)
- Hologram dev (DIDComm V2 through dev mediator) can also connect and interact with it. Since it also supports V1 and V2 by default, it should detect the V2 services and connect to it on V2 directly
- You can create an invitation/presentation-request in both V1 and V2 and Hologram dev can properly scan both of them
- You can create an invitation/credential-offer in both V1 and V2 and Hologram dev can properly scan both of them

tarunvadde:
The overall implementation is done as per the issue #427. But one open question to both @genaris and @lotharking: 

- No way to default the agent to V2-only invitations. `AGENT_DIDCOMM_VERSIONS=v2` works for the published DID doc, but the controller still requires the caller to set `didCommVersion=v2` per request, otherwise the inviter falls back to V1 handshake protocols.

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -65,6 +65,7 @@ export class InvitationController {
     return await createInvitation({
       agent: await this.agentService.getAgent(),
       useLegacyDid: options?.useLegacyDid,
+      didCommVersion: options?.didCommVersion,
```
genaris, line 68: So far I found two issue with these regular invitations:

1) If I don't set anything on AGENT_DIDCOMM_VERSIONS, it properly creates both V1 and V2 services, but if I call this endpoint without explicit `didCommVersion`, it creates a V1 invitation. Maybe this is what you've mentioned in a recent comment on this PR, but I don't understand why we have this limitation
2) When I force it to didCommV2, it works, but the resulting URL has the form `https://hologram.zone/?oob=eyJ0eXBlIjoiaHR0cHM6Ly9kaWRjb2...`. Shouldn't the query parameter be `_oob`, as per [DIDComm V2 specification](https://identity.foundation/didcomm-messaging/spec/#standard-message-encoding)?

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -201,7 +202,7 @@ export class InvitationController {
   ): Promise<CreatePresentationRequestResult> {
     const agent = await this.agentService.getAgent()
 
-    const { requestedCredentials, ref, callbackUrl, useLegacyDid } = options
+    const { requestedCredentials, ref, callbackUrl, useLegacyDid, didCommVersion } = options
```
genaris, line 207: This must be also added to credential offers, since otherwise we are blocked to V1

genaris (commented):
Some good improvements so far. However, I got a few problems I'd like to verify before merging.

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
genaris: Anyway it didn't work for me for presentation request neither. I called it like this and it created a v1 invitation:

```
curl -X 'POST' \
  'http://localhost:3000/v1/invitation/presentation-request' \
  -H 'accept: application/json' \
  -H 'Content-Type: application/json' \
  -d '{
  "ref": "1234-5678",
  "callbackUrl": "https://myhost/mycallbackurl",
  "requestedCredentials": [
    {
      "credentialDefinitionId": "did:webvh:QmSfHGyJPvfAMVsbetz3m6Q5BuyLV4TwaNsDjC1F1KLBT5:.ngrok-free.app/resources/zQmbN6aKQynYRMn4EFjvNEk89bxYT4bT62xwa3NXBsku99b",
      "attributes": [
        "phoneNumber"
      ]
    }
  ],
  "didCommVersion": "v2"
}'
```

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
tarunvadde: To answer your 2nd question, I missed it. I will raise a PR upstream to update the LINK_PARAM for V2 invitations. 

If you want a quick solution, we could do:

```
if (didCommVersion === 'v2') {
  const encoded = JsonEncoder.toBase64Url(outOfBandInvitation.toJSON())
  url = `${invitationBaseUrl}?_oob=${encoded}`
} else {
  url = outOfBandInvitation.toUrl({ domain: invitationBaseUrl })
}
```

genaris: No problem, let´s fix it on credo-ts because this also affects Hologram (and we´ll need to do it anyway).

tarunvadde: To answer your first question:

Wasn't intentional, just an unintended default-to-V1 in createInvitation. Removed it, so the endpoint now defaults to V2 when the agent supports V2.

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
tarunvadde: > This must be also added to credential offers, since otherwise we are blocked to V1

Yes, same behaviour in credential offer.

tarunvadde: > Anyway it didn't work for me for presentation request neither. I called it like this and it created a v1 invitation:
> 
> ```
> curl -X 'POST' \
>   'http://localhost:3000/v1/invitation/presentation-request' \
>   -H 'accept: application/json' \
>   -H 'Content-Type: application/json' \
>   -d '{
>   "ref": "1234-5678",
>   "callbackUrl": "https://myhost/mycallbackurl",
>   "requestedCredentials": [
>     {
>       "credentialDefinitionId": "did:webvh:QmSfHGyJPvfAMVsbetz3m6Q5BuyLV4TwaNsDjC1F1KLBT5:.ngrok-free.app/resources/zQmbN6aKQynYRMn4EFjvNEk89bxYT4bT62xwa3NXBsku99b",
>       "attributes": [
>         "phoneNumber"
>       ]
>     }
>   ],
>   "didCommVersion": "v2"
> }'
> ```

Re-tested the same curl and it now produces a v2 OOB invitation:

```
{
  "type": "https://didcomm.org/out-of-band/2.0/invitation",
  "from": "did:webvh:Qm…",
  "body": { "accept": ["didcomm/v2"] },
  "attachments": [{
    "media_type": "application/json",
    "data": { "json": { "@type": "https://didcomm.org/present-proof/2.0/request-presentation", … } }
  }]
}
```

Except the LINK_PARAM is still `oob` unless we update the deps.

genaris (approved):
Working quite well, with some remaining issues. One of them is about short URLs: created https://github.com/verana-labs/vs-agent/issues/435 to track it.

Another one is unrelated to this work but important for our testing: https://github.com/openwallet-foundation/credo-ts/pull/2788. For the moment, we keep a patch on VS Agent (it would be needed on Hologram as well until this fix is merged to Credo main and feat/didcomm-v2 branches).

PR #434, opened 2026-05-06
test: e2e vt flow at sdk level

lotharking:
**Changes**
- On going


packages/agent-sdk/src/utils/index.ts
```diff
@@ -0,0 +1,8 @@
+export * from './agent'
+export * from './data'
+export * from './setupSelfTr'
+export * from './testing'
+export * from './trustCredentialStore'
+export * from './util'
+export * from './webhook'
+export * from './webhookEvent'
```
genaris: IMHO webhooks are part of the VS Agent app. If it needs to emit specific events, VsAgent could do so (with an appropriate interface) and VS Agent app can catch them and send the webhooks based on them.

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -0,0 +1,208 @@
...
+        jsonld: {
+          credential: unsignedCredentialJson,
+          options: { proofType: 'Ed25519Signature2020', proofPurpose: 'assertionMethod' },
+        },
+      },
+    })
+    return offered
+  }
+
+  async acceptCredential(input: AcceptCredentialInput): Promise<VtFlowRecord> {
```
tarunvadde, line 169: Hey @lotharking, acceptCredential calls vtFlowApi.acceptReceivedCredential directly without verifying the offered credential first? Is this intentional?

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -0,0 +1,208 @@
...
+      throw new Error(`vt-flow record ${vtFlowRecordId} has no credentialExchangeRecordId`)
+    }
+
+    const formatData = await this.agent.didcomm.credentials.getFormatData(record.credentialExchangeRecordId)
+    const jsonld = (formatData.credential as { jsonld?: unknown } | undefined)?.jsonld
+    if (!jsonld) {
+      throw new Error(`No jsonld credential in exchange ${record.credentialExchangeRecordId}`)
+    }
+    const w3cCredential = JsonTransformer.fromJSON(jsonld, W3cJsonLdVerifiableCredential)
+    await createVtc(this.agent, this.options.publicApiBaseUrl, record.id, w3cCredential)
```
tarunvadde, line 195: Are we going to use record.id for schemaId?

packages/agent-sdk/src/utils/index.ts
lotharking: `export * from './webhook'` was removed because it is unused code.
`webhookEvent` cannot be part of the VS Agent app because it is already present in the plugin events. This was considered out of scope.

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
lotharking: Honestly, I'm not sure about it.
I prefer this kind of selector because it is unique. Before merging into `main`, please check with Ariel.

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
lotharking: Not at all. Remember, this is only a simple test.
The idea behind this test is just to guarantee that we can complete the flow. However, there are still some gaps to fully complete the spec, so please use this only as a reference.

tarunvadde: Got it! Thanks @lotharking

packages/agent-sdk/src/utils/index.ts
genaris: I know it's out of scope of this PR in particular, but this is not well architected and should be refactored at some point (ideally before shipping 1.11) if we want to trully separate SDK from the REST/web API

PR #437, opened 2026-05-08
fix: did:webvh log hashes

### LINKED ISSUE #436: Update didwebvh-ts to 2.7.4
Important to fix a critical hash calculation error in previous versions. We'll probably need to recalculate DID log after this. Any workaround for an smooth upgrade without the need of recreating wallets (and ideally maintaiing the initial SCID) would be really great. Otherwise, we'll need to recreate DIDs of all existing services.

genaris:
At the moment, the best solution I found for #436. We'll need to update Hologram to use 2.7.4 (e.g. with a new version of VerRe) before deploying all of this. So we'll need to urgently create a new release of Hologram in order to let most of our users be up to date in this sense.


packages/agent-sdk/test/logs/did_webvh_QmacbAUPqms5motYZ3yVee3Mf2Y6EMtortGKzJSTYtyaRF_example_com.jsonl
lotharking, line 1: I think it would be great to add this to the `.gitignore` file.

lotharking (approved):

PR #438, opened 2026-05-09
fix: update getShortUrl to use v1 and v2 invitations

### LINKED ISSUE #435: DIDComm invitation short URLs are always rendered as 1.1 OOB invitations
Even if we are using DIDComm V2, the invitations created for presentation-requests and credential-offers are being rendered as 1.1 OOB invitations when rendered as short URLs (i.e.  "https://public-host/s?id=f311e2c2-1f02-40a3-a522-891a9d01c2a7"). This is because /s endpoint is internally parsing the invitation URL using Credo's parseInvitation, which will always return a V1 OutOfBandInvitation object.

```
        const invitation = await agent.didcomm.oob.parseInvitation(longUrl)
        res.send(invitation.toJSON()).end()
```

This is quite important, because we almost always use the short URLs when making presentation requests and credential offers.

Changing `agent.didcomm.oob.parseInvitation` return type might need a breaking change to Credo's API, so probably another mechanism must be used (or a new Credo's OOB API method created for this). Suggestions and proposals are welcome.

tarunvadde:
As discussed in #435, `agent.didcomm.oob.parseInvitation` returns a v1 `DidCommOutOfBandInvitation` wrapper that already carries the original v2 invitation in its public `v2Invitation` field. The `/s` handler now serializes the v2 form when present, falling back to the v1 wrapper otherwise. No Credo API change needed.

### DIFF
--- a/apps/vs-agent/src/controllers/public/invitation/InvitationController.ts
+++ b/apps/vs-agent/src/controllers/public/invitation/InvitationController.ts
@@ -46,7 +46,8 @@ export class InvitationRoutesController {
           }
         }
         const invitation = await agent.didcomm.oob.parseInvitation(longUrl)
-        res.send(invitation.toJSON()).end()
+        const invitationJson = invitation.v2Invitation?.toJSON() ?? invitation.toJSON()
+        res.send(invitationJson).end()
       } else {
         res.status(302).location(longUrl).end()
       }


genaris (approved):
I think this should be catalogued as a fix rather than a mere refactor, because it indeeds fixes an issue!

PR #440, opened 2026-05-12
feat: check verifiable service requirements prior to accept vt flows

### LINKED ISSUE #428: Check Verifiable Service requirements prior to accept VT flows
We need to verify that the other party is a valid Verifiable Service before making VT-related interactions with it. Requirements are enumerated in https://verana-labs.github.io/verifiable-trust-spec/#vs-conn-vs-requirements-for-a-vs-to-accept-a-connection-from-another-service.

In order to identify the public DID related to the other party easily, we want these interactions to be done directly through DIDComm V2 and use its embedded features to associate it to a public DID:

- Applicant agent will send the first message (e.g. vt-flow's `validation-request`) directly to the public DID of the validator and using its own public DID to sign it.
- Validator agent will rotate immediately (on its initial response) to a peer DID to continue the interaction
- Applicant agent will rotate immediately (on the next message) to a peer DID to continue the interaction

lotharking:
**Changes**
- Implement DID rotation using DID Peer connections
- Replace previous connection flow with `receiveImplicitInvitation` 
- Update integration tests to support DID rotation flows
- Remove unused code
- Improve DIDComm v2 connection handling and outbound messaging flow


packages/vt-flow/src/services/VtFlowService.ts
```diff
@@ -185,7 +197,8 @@ export class VtFlowService {
   ): Promise<VtFlowRecord> {
     const { message, agentContext } = messageContext
     const connection = messageContext.assertReadyConnection()
-    await this.assertVerifiableService(agentContext, connection.id)
+    const peerDid = connection.theirDid ?? connection.id
```
tarunvadde: `connection.id` is a uuid, not a DID. If theirDid is null on the receive side, assertVerifiableService gets called with a uuid where it expects a DID.

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -45,19 +47,20 @@ export class VtFlowApi {
   }
 
   public async sendValidationRequest(options: SendValidationRequestOptions): Promise<VtFlowRecord> {
-    const connection = await this.connectionService.getById(this.agentContext, options.connectionId)
-    connection.assertReady()
-    await this.vtFlowService.assertVerifiableService(this.agentContext, connection.id)
+    const { connection } = await this.bootstrapConnection(options.recipientDid)
```
tarunvadde: This comment is for all the occurences of `bootstrapConnection`. We are not looking for existing connection here? This creates multiple records, can't we reuse a ready connection if one exists?

Just like how Hologram does it:

```
const existing = await connectionsApi.findByInvitationDid(recipientDid)
const ready = existing.find(c => c.isReady && !isTerminated(c))
if (ready) {
  return { connection: ready }
}
// otherwise, fall through to receiveImplicitInvitation
```

tarunvadde (commented):
Everything else looks good.

tarunvadde (approved):

packages/vt-flow/src/VtFlowApi.ts
genaris: I think @tarunvadde is right. `vt-flow` extension module is responsible only for managing vt-flow protocol instances. The connection itself at DIDComm protocol level is out of its scope.

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -277,11 +285,51 @@ export class VtFlowApi {
...
+    const oobApi = this.agentContext.dependencyManager.resolve(DidCommOutOfBandApi)
+    const connectionsApi = this.agentContext.dependencyManager.resolve(DidCommConnectionsApi)
+    const didRepository = this.agentContext.dependencyManager.resolve(DidRepository)
+
+    const existing = await connectionsApi.findByInvitationDid(recipientDid)
+    const ready = existing.find(c => c.isReady)
+    if (ready) return { connection: ready }
+
+    const [createdDidRecord] = await didRepository.getCreatedDids(this.agentContext, { method: 'webvh' })
+    const ourDid = createdDidRecord?.did
```
genaris: This is already known by VsAgent object and vs-agent app, so all of this logic is not needed.

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -277,11 +285,51 @@ export class VtFlowApi {
...
+    const connection = await connectionsApi.getById(connectionRecord.id)
+    return { connection }
+  }
+
+  private async ensureRotated(record: VtFlowRecord): Promise<void> {
+    if (record.hasRotated) return
+    await this.vtFlowService.rotateAndWait(this.agentContext, record.connectionId)
+    record.hasRotated = true
+    await this.vtFlowService.updateRecord(this.agentContext, record)
+  }
```
genaris: Rotation is managed automatically by Credo under the hood. You should not worry about it (neither be interested on the generated peer DID). The only thing you should care about (and not within vt-flow module, but at VsAgent or vs-agent app level) is to check whether an existing connection related to the another party DID ("recipientDid") and use the connectionId of it to initiate the vt-flow protocol.

genaris (changes requested):
I don't think this PR (which by the way IS NOT a fix) is addressing what it claims to be. I would expect it to implement `assertVerifiableService` hook, where the actual VS requirements would be checked (e.g. by using VPR trust resolution endpoints), and instead it is reimplementing some DID rotation logic that is (or should be) actually embedded directly into credo-ts DIDComm connection establishment logic.

But the VS-CONN-VS implementation alone might not be significant enough to raise a PR at the moment, as it should be wired with some real use case so it can be meaningful and tested. So I think this PR should be completed after the basic onboarding process flow ([VSA-VTI-FLOW-OP-NEW](https://github.com/verana-labs/verana-spec/blob/main/v4/vs-agent/spec.md#vsa-vti-flow-op-new-new-onboarding-process)) is implemented.

PR #442, opened 2026-05-13
feat(vt-flow): applicant verification + on-chain permission session

tarunvadde:
- New `verifyOfferedCredential` runs the on-chain validator-active check , plus a digest soft check, wired into the e2e `verifyCredential` hook so the verification path is actually utilized.
- `validateAndOfferCredential` now calls `createOrUpdatePermissionSession` after `setPermissionVPToValidated`
- `publishCredentialAsLinkedVp` extracts `schemaBaseId` from `credentialSchema.id` so the LinkedVP endpoint matches the schema.


packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -171,9 +183,54 @@ export class VtFlowOrchestrator {
...
+    const vtFlowApi = this.resolveVtFlowApi()
+    const record = await vtFlowApi.findById(vtFlowRecordId)
+    if (!record) throw new Error(`vt-flow record ${vtFlowRecordId} not found`)
+    if (record.role !== VtFlowRole.Applicant) throw new Error('Record is not applicant-side')
+    if (!record.permId) throw new Error('Record has no permId')
+    if (!record.credentialExchangeRecordId) {
+      throw new Error('Record has no credentialExchangeRecordId; nothing to verify')
+    }
+
+    const holderPerm = await chain.getPermission(Number(record.permId))
```
lotharking: If possible, I would prefer to use the indexer API to resolve this.
However, we should keep in mind that there is a possibility that the indexer may become inconsistent with the ledger. In those cases, I would rather detect and surface these kinds of errors explicitly to preserve consistency.

lotharking (commented):
Just a suggestion: please don’t forget to run this test with a funded wallet

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
tarunvadde: Switched the two perm lookups in `verifyOfferedCredential` to `VeranaIndexerService.getPermission()`. Kept chain RPC as a fallback when callers don't wire an indexer (auto-flow path, existing tests) so this doesn't break those. Please let me know if there's anything else to be done here.

lotharking (approved):

PR #444, opened 2026-05-14
feat: better filtering in getAllConnections API

tarunvadde:
Extend `GET /v1/connections` with additional filters to make the endpoint more useful in V2 flows.


lotharking (approved):

genaris (approved):
Can you bump credo-ts pr-2704 version once it is released so the query by didcommVersion works?

Issue #417, opened 2026-04-22
Credential revocation status in presentations and presentation-request callback


genaris:
GET /v1/presentations is giving us most details of proof exchange flows, but it is unclear to me how it responds when a presented proof is revoked. It is theoretically possible to present a revoked credential (as long as we don't request a specific revocation status in the proof request (which we don't).

Same thing happens to the callback for /v1/invitations/presentation-request, as we provide a presentation status and claims but we don't even give a `verified` field. 

So the goal of this issue is investigate what happens when we request a credential and Hologram has a revoked one in its wallet. We know Hologram might not be aware of this revocation status (since we are not fully handling revocation notification, and we don't send it in all VS Agent flows), so I don't know what happens in this situation, since Hologram might want to create a Non-revocation proof and it will of course fail to do so.

Check [AnonCreds specification](https://anoncreds.github.io/anoncreds-spec/) for more information about the concepts and the proof request flow.

tarunvadde:
## Investigation: where verification leaks past revocation, and where it would silently break if we naively fix it

I dug into VS Agent, Credo-TS source, the AnonCreds spec, and the Hologram source. Findings below, with file:line references throughout.

### What's happening today

#### 1. VS Agent never asks for a non-revocation proof

The proof request is built at `apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts:282-287`:

```ts
await agent.didcomm.proofs.createRequest({
  protocolVersion: 'v2',
  proofFormats: {
    anoncreds: { name: 'proof-request', version: '1.0', requested_attributes: requestedAttributes },
  },
})
```

No `non_revoked` field. Credo accepts this as "no revocation interval requested" (`AnonCredsProofRequest.non_revoked?` is optional).

#### 2. Verification runs without consulting any revocation registry

Credo's `AnonCredsRsVerifierService.verifyProof` (`credo-ts/packages/anoncreds/src/anoncreds-rs/AnonCredsRsVerifierService.ts:20-69`) only loops over `requestedNonRevokedRestrictions`, which is empty when the request had no `non_revoked`. `revocationRegistries` is empty in the call to anoncreds-rs. The proof's signature, schema, and cred-def are checked. Revocation is not.

Result: `record.isVerified = true` for valid signatures, regardless of revocation.

#### 3. The GET endpoint exposes `verified`; the callback does not

`GET /v1/presentations` returns `verified: proofExchange.isVerified ?? false` (`PresentationsController.ts:146`).

[...]

PR #445, opened 2026-05-19
feat(vs-agent): proof non-revocation support

tarunvadde:
Based on #417

## Summary

- `createPresentationRequest` now accepts `requireNonRevocation` (defaults true). Threads a `non_revoked` interval into credo's anoncreds proof formats and emits the resulting `verified` flag in the presentation callback.
- `revokeCredential` now persists the new status list as an AttestedResource and appends it to the parent rev-reg-def's `links[]`. Without this, verifiers resolving via did:webvh never saw revocations. Although I saw a similar implementation for `sendMessage` at `MessageController.ts`, not sure why it's separate. If this was intentional please let me know if I need to revert back here.
- did:webvh DID documents no longer leak the didwebvh-ts update key into `authentication`. credo's DidCommMessageSender was picking it as the sender VM, producing a non-spec-compliant DIDComm v2 skid. Mirrors didcomm-mediator's REPLACE pattern; one-shot migration cleans up already-published records.
- New `GET /v1/credentials` admin endpoint listing credential exchanges with AnonCreds metadata + attributes.
- Bumps credo to `0.7.1-pr-2704-20260519192541`; drops the local webvh timestamp patch (now upstream as PR #2788).

## Scenarios Tested

- Issue a revocable VC to Hologram; Request NRP, Hologram presents proof, the proof presentation is `verified: true`. (both DIDComm V1 & V2)
- Revoke the VC; Request NRP, Hologram presents proof, the proof presentation is `verified: false and state: abandoned`.(both DIDComm V1 & V2)


apps/vs-agent/src/controllers/admin/credentials/dto/credential-exchange-data.dto.ts
```diff
@@ -0,0 +1,60 @@
+import { ApiProperty } from '@nestjs/swagger'
+import { Claim } from '@verana-labs/vs-agent-model'
+
+export class CredentialExchangeDataDto {
```
lotharking, line 5: You should include `class-validator` to control the typing.

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,114 @@
...
+    }
+
+    try {
+      return await this.toDto(record)
+    } catch (error) {
+      throw new InternalServerErrorException(error)
+    }
+  }
+
+  private async toDto(record: DidCommCredentialExchangeRecord): Promise<CredentialExchangeDataDto> {
```
lotharking: Receive the agent to avoid inconsistencies
```suggestion
  private async toDto(agent: VSAgent, record: DidCommCredentialExchangeRecord): Promise<CredentialExchangeDataDto> {
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,114 @@
...
+
+  @Get('/:credentialExchangeId')
+  @ApiOperation({ summary: 'Get credential exchange by id' })
+  @ApiOkResponse({ description: 'Credential exchange data', type: CredentialExchangeDataDto })
+  @ApiBadRequestResponse({ description: 'Invalid credentialExchangeId' })
+  @ApiNotFoundResponse({ description: 'Credential exchange not found' })
+  public async getById(
+    @Param('credentialExchangeId') credentialExchangeId: string,
+  ): Promise<CredentialExchangeDataDto> {
+    if (!credentialExchangeId) {
```
lotharking: If you include `class validator` this should be removed.

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,114 @@
...
+    let offerAttributes: Claim[] | undefined
+    try {
+      const formatData = await agent.didcomm.credentials.getFormatData(record.id)
+      if (formatData.offerAttributes?.length) {
+        offerAttributes = formatData.offerAttributes.map(
+          attr => new Claim({ name: attr.name, value: attr.value, mimeType: attr.mimeType }),
+        )
+      }
+    } catch (error) {
+      this.logger.debug(`Could not load format data for ${record.id}: ${error}`)
```
lotharking: `error` should be [object Object]
```suggestion
      this.logger.debug(`Could not load format data for ${record.id}: ${JSON.stringify(error)}`)
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,114 @@
...
+    summary: 'List all credential exchanges',
+    description:
+      'Returns every credential exchange record tracked by VS Agent. Useful for inspecting the issuance pipeline during testing.',
+  })
+  @ApiOkResponse({
+    description: 'Array of credential exchange data',
+    type: CredentialExchangeDataDto,
+    isArray: true,
+  })
+  public async getAll(): Promise<CredentialExchangeDataDto[]> {
```
lotharking: Include try catch to prevent errors

lotharking (changes requested):

apps/vs-agent/src/controllers/admin/verifiable/TrustService.ts
```diff
@@ -293,6 +295,52 @@ export class TrustService {
           if (!uptStatusListResult.revocationStatusListState.revocationStatusList) {
             throw new Error(`Failed to update revocation status list`)
           }
+
```
lotharking, line 296: Please review this logic. There is a similar implementation in the credential definition flow, and it would be better to refactor this into a dedicated reusable method.

tarunvadde: Done. Created a helper `CredentialTypesService.appendStatusListToRevocationRegistry` so both
callers share one code path.

lotharking: Do you check the `revRegDefRecord`? Can this be refactored?

lotharking (commented):
Hey Tarum only a comment

genaris (changes requested):
In general I think it works great, but I'd like to make `requireNonRevocation` false by default, just to maintain compatibility with existing Hologram deployments (otherwise, I think they will always fail).

BTW have you checked what happens if we set `requireNonRevocation` to true in case of non-revocable credentials? IIRC, Credo just ignores it in case the underlying credential definition does not support revocation, but it might be a good idea to do a quick E2E test on this use-case just to make sure nothing unexpected happens.

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
genaris, line 1: Make sure to mention this new requireNonRevocation flag in the OpenAPI docs, so it is available in Swagger UI

apps/vs-agent/src/controllers/admin/credentials/dto/credential-exchange-data.dto.ts
lotharking, line 1: Remember to update the `Readme.md` about this controller and the new parameter.

apps/vs-agent/src/controllers/admin/credentials/dto/credential-exchange-data.dto.ts
```diff
@@ -0,0 +1,80 @@
...
+    type: [Object],
+    required: false,
+    example: [
+      { name: 'firstName', value: 'Alice' },
+      { name: 'age', value: '30' },
+    ],
+  })
+  @IsOptional()
+  @IsArray()
+  offerAttributes?: Claim[]
```
lotharking: For consistency with the current codebase, this should always be named `claims`
```suggestion
  claims?: Claim[]
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,123 @@
...
```
lotharking: Maybe this should be named something like `getAllCredentialExchanges`.

```suggestion
  public async getAllCredentialExchanges(): Promise<CredentialExchangeDataDto[]> {
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,123 @@
...
+      throw new InternalServerErrorException(error)
+    }
+  }
+
+  @Get('/:credentialExchangeId')
+  @ApiOperation({ summary: 'Get credential exchange by id' })
+  @ApiOkResponse({ description: 'Credential exchange data', type: CredentialExchangeDataDto })
+  @ApiBadRequestResponse({ description: 'Invalid credentialExchangeId' })
+  @ApiNotFoundResponse({ description: 'Credential exchange not found' })
+  public async getById(
```
lotharking: Maybe this should be named something like `getCredentialExchangeById`.
```suggestion
  public async getCredentialExchangeById(
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,123 @@
...
+    }
+
+    try {
+      return await this.toDto(agent, record)
+    } catch (error) {
+      throw new InternalServerErrorException(error)
+    }
+  }
+
+  private async toDto(
```
lotharking:
```suggestion
  private async getCredentialExchangeData(
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,123 @@
...
+import { CredentialExchangeDataDto } from './dto/credential-exchange-data.dto'
+
+interface AnonCredsCredentialMetadata {
+  credentialDefinitionId?: string
+  schemaId?: string
+}
+
+@ApiTags('credentials')
+@Controller({
+  path: 'credentials',
```
lotharking:
```suggestion
  path: 'credential-exchanges',
```

apps/vs-agent/src/controllers/admin/credentials/CredentialExchangesController.ts
```diff
@@ -0,0 +1,123 @@
...
+import { VsAgentService } from '../../../services/VsAgentService'
+
+import { CredentialExchangeDataDto } from './dto/credential-exchange-data.dto'
+
+interface AnonCredsCredentialMetadata {
+  credentialDefinitionId?: string
+  schemaId?: string
+}
+
+@ApiTags('credentials')
```
lotharking:
```suggestion
@ApiTags('credential-exchanges')
```

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
tarunvadde: Added. 

Also I already confirmed via E2E test against a non-revocable credential definition with requireNonRevocation: true in the request and the expected behaviour is credo silently skips the NRP.

lotharking (approved):

genaris (approved):

PR #446, opened 2026-05-19
feat(agent): listen to DidCommConnectionDidRotated event so that we can listen to termination status of didcomm v2 connection

tarunvadde:
Listen to DidCommConnectionDidRotated event so that we can listen to termination status of didcomm v2 connection


apps/vs-agent/src/events/ConnectionEvents.ts
```diff
@@ -109,6 +110,26 @@ export const connectionEvents = async (agent: VsAgent<any>, config: ServerConfig
...
+      const body = new ConnectionStateUpdated({
+        connectionId: record.id,
+        invitationId: record.outOfBandId,
+        state: 'terminated',
+      })
+
+      await sendWebhookEvent(config.webhookUrl + '/connection-state-updated', body, config.logger)
+    },
+  )
+
```
genaris: Is it possible to merge both? IIRC, this event has been added (relatively) recently and should work for both. And if not... it would be a good idea to do so!

tarunvadde: Good call, done. Merged and removed the separate DidCommHangupMessage listener. Verified end-to-end against Hologram for both versions.

genaris (approved):

PR #447, opened 2026-05-21
feat: decouple webhook event delivery from sdk

lotharking:
**Changes**
- Remove legacy webhook file
- Create `EventEmitter` class to handle events
- Replace `events` with `webhookUrl` configuration 
- Update testing


packages/agent-sdk/src/utils/EventEmitter.ts
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * Single entry point that plugins use to publish events. It is polymorphic on its argument:
+ *  - an {@link Event} (e.g. MessageStateUpdated, ConnectionStateUpdated) is delivered as-is
+ *  - a raw domain {@link BaseMessage} is wrapped into a {@link MessageReceived} event
+ *
+ * The concrete delivery mechanism is the {@link EventSink} injected by the app, keeping
+ * plugins decoupled from webhooks/HTTP.
+ */
+export class EventEmitter {
```
tarunvadde: This class is named `EventEmitter`, which collides with Credo's `EventEmitter` from `@credo-ts/core`. So we will have to use alias in order to import both if needed. Would that be ok? Just asking

apps/vs-agent/src/main.ts
```diff
@@ -231,7 +233,7 @@ const run = async () => {
     port: ADMIN_PORT,
     cors: USE_CORS,
     logger: serverLogger,
-    webhookUrl: EVENTS_BASE_URL,
+    events: new EventEmitter(webhookEvent(EVENTS_BASE_URL, serverLogger)),
```
tarunvadde: This is constructed unconditionally even if `EVENTS_BASE_URL` is unset/empty. The old `webhookUrl?: string` was optional in `ServerConfig`, now `events: EventEmitter` is required. So every emit attempts a malformed POST when no webhook is configured.

packages/agent-sdk/src/types.ts
```diff
@@ -4,12 +4,13 @@ import { BaseLogger } from '@credo-ts/core'
...
 
 import { BaseAgentModules, VsAgent } from './agent/VsAgent'
+import { EventEmitter } from './utils/EventEmitter'
 
 export const MESSAGE_HANDLERS = 'MESSAGE_HANDLERS'
 
 export interface VsAgentPluginConfig {
   logger: BaseLogger
-  webhookUrl?: string
+  events: EventEmitter
```
tarunvadde: Same here. `events: EventEmitter` is now required where `webhookUrl?: string` was optional. This breaks any external plugin author who constructed their own `VsAgentPluginConfig`.

tarunvadde (commented):
Won't this PR conflict with #448?

packages/agent-sdk/src/utils/EventEmitter.ts
lotharking: You are completely right, thanks. I used `EventPublisher` instead of `EventEmitter`.

apps/vs-agent/src/main.ts
lotharking: You are right. However, even in the previous implementation, we already had a default value for `EVENTS_BASE_URL`. When it is not set correctly, the application already throws an error like:
```
`Error sending ${body.type} webhook event to ${webhookUrl}`
```

packages/agent-sdk/src/types.ts
lotharking: You are right. The agent is not only used for events, so this could become a headache. Using an optional field instead of a required one makes more sense here.

tarunvadde (approved):

packages/agent-sdk/src/utils/EventPublisher.ts
genaris, line 1: It's hard to me to find value on this class. Why don't we just use VsAgent's EventEmitter (the one that comes from credo-ts), defining a few custom events (e.g `vs-agent-connection-state-updated`, `vs-agent-message-received`. `vs-agent-message-state-updated`) and subscribe on the app to execute the corresponding webhooks (or do whatever we need)?

genaris (changes requested):
I think the SDK (with its VsAgent object) and its plug-ins should be the ones in charge of emitting events, and the app being the one that listens for them and executes the logic it requires (logging, webhooks, executing callbacks, etc.).

My first suggestion is to simplify the implementation by using VsAgent's internal emitter, and a second one (and bigger in terms of refactoring) would be to move most of the credo-ts message processing (i.e. most of what is currently under `apps/vs-agent/BaseMessageEvents` and `apps/vs-agent/ConnectionEvents`) to SDK, in such a way that it serves as an abstraction layer of any complexity given by DIDComm/credo-ts's internal events. In my opinion, ideally the SDK should trigger a `MessageReceived` event and vs-agent app listen to it and trigger webhooks (or executing presentation callback in case of presentations).

With these two together, I think this refactoring will be significant enough to have a consistent behaviour in terms of event flow.

apps/vs-agent/src/plugins/MessagingPlugin.ts
genaris: We are still making the app responsible for VsAgent subscription to credo-ts events, so the VsAgent specific events will not be triggered unless the application does this manual call to `baseMessageEvents`.

I'd expect VsAgent to internally subscribe to credo-ts events (e.g. at the end of its `initialize()` method) and trigger its own events, so apps just need to subscribe to them as needed.

genaris (changes requested):

packages/agent-sdk/src/events/VsAgentEvents.ts
```diff
@@ -0,0 +1,64 @@
...
+  const event =
+    eventOrMessage instanceof Event
+      ? eventOrMessage
+      : new MessageReceived({ timestamp: eventOrMessage.timestamp, message: eventOrMessage })
+
+  agent.events.emit(agent.context, {
+    type: busTypeByEventType[event.type],
+    payload: { event },
+  })
+}
```
genaris: I don't see the reason of making this so ugly, i.e. just having a method called `emitVsAgentEvent(agent: VsAgent, event: Event)` and a separate `emitVsAgentMessageReceivedEvent(agent: VsAgent, message: BaseMessage)`. Or even better: don't create any utility method at all, and just call e.g. `agent.events.emit(agent.context, { type: VsAgentEventType.ConnectionStateUpdated, payload: { event }` in the method that is actually sending such a message.

Why don't simply follow the event pattern we have in Credo, in such a way we won't need to do these "bus channel names"? I think this is overcomplicating the things.

apps/vs-agent/src/utils/webhookEvent.ts
```diff
@@ -0,0 +1,29 @@
...
+      })
+    } catch (error) {
+      logger.error(`Error sending ${body.type} webhook event to ${webhookUrl}`, {
+        cause: error,
+      })
+    }
+  }
+
+  for (const type of Object.values(VsAgentEventType)) {
+    agent.events.on(type, ({ payload }: VsAgentEvent) => post(payload.event as Event))
```
genaris: You are not forced to make VsAgent SDK events look exactly the same as VS Agent app does for webhooks. You can respect the current API and do the corresponding transformation right here, instead of forcing the SDK to adapt to this specific need.

packages/agent-sdk/src/events/BaseMessageEvents.ts
```diff
@@ -17,11 +18,13 @@ import {
   IdentityProofSubmitMessage,
   VerifiableCredentialSubmittedProofItem,
 } from '@verana-labs/vs-agent-model'
-import { getRecordId, sendMessageReceivedEvent, VsAgent } from '@verana-labs/vs-agent-sdk'
+
+import { getRecordId } from '../utils/agent'
 
 import { PresentationStatus, sendPresentationCallbackEvent } from './CallbackEvent'
```
genaris: I don't like the fact that the current code of  `sendPresentationCallbackEvent` is executed right at the SDK side. I think it is a great example of a new VS Agent event type, such as `vs-agent-presentation-status-updated` (or something consistent with current events for messages and connections), that will be catched by VS Agent app and, if a callbackUrl is defined, will do the appropriate `fetch` sending all the data as required.

So within `sendPresentationCallbackEvent`, instead of having the fetch we can have the `emitVsAgentEvent` call (or simply an agent.events.emit(xxx, yyy) and place the `fetch` directly within a listener in apps/vs-agent.

Of course, the file instead of being called "CallbackEvent" it will be "PresentationEvents" and named appropriately. This is important because at some point in the near future we'll do the same with credential offers (we didn't implement callback URLs for that yet).

genaris (changes requested):
It's looking better, but I think it still has some overengineered pieces that can be simplified.

packages/agent-sdk/src/events/VsAgentEvents.ts
lotharking: Got it. I created a single `emitVsAgentEvent` function mainly to avoid duplicating the emit logic, but I agree that conceptually it should stay aligned with the Credo event structure.

packages/model/src/events/EventType.ts
```diff
@@ -2,4 +2,5 @@ export enum EventType {
   ConnectionState = 'connection-state-updated',
   MessageStateUpdated = 'message-state-updated',
   MessageReceived = 'message-received',
+  PresentationStatusUpdated = 'presentation-status-updated',
```
genaris:
```suggestion
ConnectionStateUpdated = 'connection-state-updated',
MessageStateUpdated = 'message-state-updated',
MessageReceived = 'message-received',
  PresentationStateUpdated = 'presentation-state-updated',
```

This makes event types consistent in naming.

genaris (approved):
LGTM, just a comment to make event naming consistent.

PR #448, opened 2026-05-21
feat: create vt flow events

### LINKED ISSUE #429: Events interface for vt-flow
We have to consider emitting webhooks events whenever vt-flow status change, so VS Agent controllers are aware of updates and can act in consequence. TBD the details of their contents, but we can follow the same pattern as we do for connections and messages.

lotharking:
**Changes**
- Create `vtFlow` update events.
- Create `VtFlowStateUpdated` message.
- Include `vtFlow` by default on VS Agent SDK.
- Update `vtFlow` testing with events.


apps/vs-agent/package.json
```diff
@@ -96,6 +95,7 @@
     "@types/multer": "^1.4.7",
     "@types/node-fetch": "^2.6.4",
     "@types/uuid": "^8.3.4",
+    "@verana-labs/credo-ts-didcomm-vt-flow": "workspace:*",
```
tarunvadde, line 98: Why did we move @verana-labs/credo-ts-didcomm-vt-flow moved to devDependencies? Will this not break the other things such as docker build?

packages/agent-sdk/src/events/VtFlowEvents.ts
```diff
@@ -0,0 +1,31 @@
...
+      state: payload.state,
+      previousState: payload.previousState,
+      permId: record.permId,
+      schemaId: record.schemaId,
+      claims: record.claims,
+      credentialExchangeRecordId: record.credentialExchangeRecordId,
+      timestamp: record.updatedAt,
+    })
+
+    await sendWebhookEvent(config.webhookUrl + '/' + EventType.VtFlowStateUpdated, body, config.logger)
```
tarunvadde: If `config.webhookUrl` ends with `/` you'll POST to `https://..//vt-flow-state-updated`, and if it's unset (sicne it's optional) you'll POST to `"undefined/vt-flow-state-updated"`. Guard with `if (!config.webhookUrl) return` and use `new URL(EventType.VtFlowStateUpdated, config.webhookUrl).toString()` or a trim.

packages/model/src/events/VtFlowStateUpdated.ts
```diff
@@ -0,0 +1,99 @@
...
+  @Expose()
+  @IsOptional()
+  @IsObject()
+  public claims?: Record<string, unknown>
+
+  @Expose()
+  @IsOptional()
+  @IsString()
+  public credentialExchangeRecordId?: string
+}
```
tarunvadde, line 127: The `VtFlowRecord` has four fields that subscribers will want but are dropped here: `errorMessage`, `subprotocolThid, `agentPermId` and `walletAgentPermId`. Suggest adding all four to options + class body, all `@IsOptional() @IsString()`.

tarunvadde (changes requested):

packages/agent-sdk/src/events/VtFlowEvents.ts
lotharking: control over input constant

tarunvadde (approved):
LGTM

genaris (approved):

PR #457, opened 2026-06-15
feat(vt-flow): align with onboarding-process spec and remove legacy one

tarunvadde:
Aligns the `vt-flow` package with the v4 [spec](https://github.com/verana-labs/verana-spec/blob/main/v4/vt-flow-protocol/spec.md). 

Closes #456.


packages/vt-flow/src/messages/OnboardingRequestMessage.ts
```diff
@@ -0,0 +1,65 @@
...
+import { IsArray, IsInstance, IsObject, IsOptional, IsString, IsUUID, ValidateNested } from 'class-validator'
+
+import { VT_FLOW_ONBOARDING_REQUEST_TYPE } from './VtFlowProtocol'
+
+export interface OnboardingRequestMessageOptions {
+  id?: string
+  participantId: string
+  participantSessionId: string
+  agentParticipantId: string
+  walletAgentParticipantId: string
```
genaris, line 12: I guess we should wait until https://github.com/verana-labs/verana/issues/340 is fixed/clarified before merging this, right?

tarunvadde: These are REQUIRED as per our vt-flow-protocol spec, and independent of https://github.com/verana-labs/verana/issues/340. I think we can merge this because the message field doesn't feed the chain session, so [#340](https://github.com/verana-labs/verana/issues/340) can't affect it.

genaris (approved):

PR #462, opened 2026-06-23
refactor(agent-sdk): serve tails by content hash, drop in-memory index

tarunvadde:
Follow-up to #459

The old `FullTailsFileService` was basically the credo-ts sample: it saved each tails file under a random uuid and used an `index.json` plus an in-memory map to look it up. That map only lives in one process, so with more than one instance a tails URL written by one returns 404 on the others.

This makes it simpler and multi-instance safe:

- Name each tails file by its content hash and serve it straight from disk (no uuid, no `index.json`, no in-memory map).
- Read the directory from Credo's data path, with `TAILS_DIRECTORY_PATH` to point it at a shared volume across instances.
- Only allow a plain file name in the URL, to avoid path traversal.

Tested with one instance, and with two instances sharing a folder where one writes the tails file and the other serves it.


packages/agent-sdk/src/credentials/FullTailsFileService.ts
lotharking, line 1: I have a question about legacy support. Do we still need it? Will legacy credentials have any issues with this implementation?

tarunvadde: By "legacy" I take it you mean registries created with the current uuid-based implementation. Those would break with this PR: the tails URL now ends in the content hash instead of a uuid, and the new handler only accepts a plain hash, so an old uuid URL returns 404. For did:webvh the published tailsLocation is fixed once published, so we can't rewrite it, the registry would have to be re-created.

In practice the uuid scheme only landed in #459, so there shouldn't be many (if any) real registries on it yet, and the demo one was already broken. So I'd lean towards not adding legacy support and re-creating any affected registry.

If there are deployed registries we must keep working, I can add a small fallback: when the id is an old uuid, look it up in the previous index.json and serve the old file. That brings back a bit of the old code, so I'd only do it if we actually need it. What do you think?

genaris: Yeah I think it would be nice if we can support previously created tail files, as there are a few partners that have been using it (not yet in production environment AFAIK, so if it becomes too hard, we can discuss with them about it or just mark this as a breaking change).

tarunvadde: Done, no breaking change needed. On the first request for an old uuid tails URL, the agent finds the previously created file and copies it into the durable dir, so the old URLs keep resolving. It checks every location older versions used: the original `./tails`, `~/.afj/tails`, and `TAILS_DIRECTORY_PATH` if set.

packages/agent-sdk/src/credentials/FullTailsFileService.ts
```diff
@@ -1,96 +1,102 @@
...
 }
-export const tailsIndex = (
-  fs.existsSync(indexFilePath) ? JSON.parse(fs.readFileSync(indexFilePath, { encoding: 'utf-8' })) : {}
-) as Record<string, string>
 
-function fileHash(filePath: string, algorithm = 'sha256') {
-  return new Promise<string>((resolve, reject) => {
-    const shasum = createHash(algorithm)
+// Copy uuid tails files from earlier locations (TAILS_DIRECTORY_PATH, ~/.afj/tails, the old cwd ./tails) into the current dir so URLs published before the hash-based change still resolve.
+export function migrateLegacyTailsFiles(agentContext: AgentContext): void {
```
lotharking, line 25:
```suggestion
// TODO (last legacy version: v1.12.0-dev.8): Remove this once the last release using the legacy configuration is no longer supported.
export function migrateLegacyTailsFiles(agentContext: AgentContext): void {
```

genaris (approved):

Issue #419, opened 2026-04-27
[VPR] Call MOD-PERM-MSG-15 TriggerResolver when public W3C credentials are added or updated in the DID Document


mjfelis:
## Context

A new Msg `MOD-PERM-MSG-15 TriggerResolver` is being added to the Verana ledger:

- VPR spec PR: https://github.com/verana-labs/verifiable-trust-vpr-spec/pull/125 (spec v4-draft14)
- Ledger implementation issue: https://github.com/verana-labs/verana/issues/287
- Indexer integration issue: https://github.com/verana-labs/verana-indexer/issues/224
- Related VS-Agent integration work: #402 ([VPR] Verana VPR Integration)

This Msg does **not** modify any VPR state. It emits an on-chain event that tells the Verana indexer to **re-run trust resolution** for the DID registered in `Permission[perm_id]`. Without this signal, the resolver will not pick up new or updated credentials published in a VS's DID Document, and the service will not become visible (or up-to-date) in the indexer's trust data.

Per the Verifiable Trust spec (`verifiable-trust-spec/spec.md`, section `[TR] Trust Resolution` and the "Trust Resolution Examples" section), trust resolution happens **off-chain** against indexed artifacts — DID Documents, linked-VPs, VTCs, and VPR records. The resolver does not observe DID Document mutations on its own; it depends on an explicit signal to know that it should resolve a given DID again.

## Required VS-Agent behavior

A VS-Agent running as a `vs_operator` for a permission whose `perm.did` equals the agent's DID SHOULD call `MsgTriggerResolver` (with that `perm.id`) in the following cases:

1. **First publication of public W3C credentials in the DID Document.** As soon as the VS-Agent publishes the `linked-vp` entry in its DID Document containing one or more public **W3C Verifiable Trust Credentials** (see `[VT-CRED-W3C]`), it SHOULD send `MsgTriggerResolver` so the resolver picks up the service and marks it as `TRUSTED` / visible.
2. **Any subsequent update to public W3C credentials in the DID Document.** This includes: a credential is replaced (new issuance), removed (revocation handled off-chain, or credential expiry), the DID Document linked-vp entries are re-generated, or any additional public W3C credential is added.
3. **Any update to the list of public credentials published via linked-vp.** Even if the underlying credential payload is unchanged, if the set of credentials surfaced in the DID Document changes, a new resolution MUST be triggered.

For clarity, the following do NOT in themselves require a `MsgTriggerResolver` call from the agent:

[...]

PR #468, opened 2026-06-25
feat(vs-agent): service endpoint management admin API and trigger resolver

tarunvadde:
Adds the Service Endpoint Management admin API (VSA-ADM-SE, #465) and the on-chain TriggerResolver it calls after a DID Document change (#419).

## Service endpoint management (#465)
Admin module for the consumable service entries in the agent's DID Document (MCP, A2A, LinkedDomains, etc.):
- GET/POST `/v1/vt/service-endpoints`, PATCH/DELETE `/v1/vt/service-endpoints/{id}`.
- Guards the auto-managed types (DIDComm v2, LinkedVerifiablePresentation, VsAgentAdminAPI) with the spec error codes; validates `serviceEndpoint` per DID-CORE.
- Each successful mutation republishes the DID Document and triggers a resolver refresh.

## Trigger resolver (#419)
After a DID Document change the agent sends `MsgTriggerResolver` (MOD-PP-MSG-15) so the Verana resolver re-evaluates the service:
- `VeranaChainService.triggerResolver`, signed by the agent's `vs_operator`.
- The chain only allows a `vs_operator` self-trigger (Path 1) for a HOLDER participant, so the hook selects the agent's active HOLDER participant by DID and skips when there is none.
- Best-effort, behind `VERANA_AUTO_TRIGGER_RESOLVER` (default true).

## Tests
- Adds a HOLDER trigger e2e (RUN_FLOW_E2E): it builds a real ECOSYSTEM -> ISSUER -> HOLDER chain and triggers the resolver via the HOLDER's own `vs_operator`. Verified locally on the CI image.
- The existing dual-operator onboarding e2e is unchanged.

Closes #465. Part of #419 (the TriggerResolver primitive and the service-endpoint trigger path).


packages/agent-sdk/tests/e2e/onboarding-dual-operator.e2e.test.ts
```diff
@@ -109,4 +114,59 @@ describeE2E('vt-flow onboarding chain integration: dual-operator validate + 0/0
...
+      const schema = await chainA.createCredentialSchema(corp.policyAddress, {
+        ecosystemId: eco.ecosystemId,
+        jsonSchema: MINIMAL_SCHEMA,
+      })
+      const root = await chainA.createRootParticipant(corp.policyAddress, {
+        schemaId: schema.schemaId,
+        did: `did:example:eco-root-${suffix}`,
+      })
+
+      // HOLDER onboarding requires an ISSUER validator (chain holder_onboarding_mode = ISSUER_VALIDATION).
```
lotharking, line 135: Review the verbose

packages/agent-sdk/src/blockchain/types.ts
```diff
@@ -196,6 +196,8 @@ export interface VeranaChainConfig {
   logger: BaseLogger
   gasPrice?: string
   corporationAddress?: string
+  /** Auto-send TriggerResolver after DID Document changes. Default true. */
```
lotharking: Too much verbose

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
lotharking, line 1: Why create a service endpoint utils and not a service?

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsController.ts
```diff
@@ -0,0 +1,116 @@
...
+    }
+  }
+
+  private async requireAgentWithDid(): Promise<VsAgent> {
+    const agent = await this.agentService.getAgent()
+    if (!agent.did) throw new BadRequestException('Agent has no public DID')
+    return agent
+  }
+
+  private toHttpException(error: unknown): Error {
```
lotharking: Why not create a global HTTP exception handler?

packages/agent-sdk/tests/e2e/onboarding-dual-operator.e2e.test.ts
tarunvadde: I thought this was needed for clarity. Removing this.

packages/agent-sdk/src/blockchain/types.ts
tarunvadde: Same here.

tarunvadde: I will remove `/** Auto-send TriggerResolver after DID Document changes. Default true. */` but keep the FIXME as is.

packages/agent-sdk/tests/e2e/onboarding-dual-operator.e2e.test.ts
tarunvadde: Actually, these comments document chain constraints that aren't visible in the test code (onboarding mode, the digest-null rule, the dual-operator authz split), so the test reads as a reference for writing other chain e2e tests. I'd recommend keeping them (atleast till the completion as per V4 spec/removing the workaround with dual operators), but happy to tighten the wording if you feel strongly.

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
tarunvadde: This lives in the SDK so a NestJS @Injectable service doesn't fit here without coupling the library to Nest. I'm not sure what to do here, could you please elaborate on what I can do better?

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsController.ts
tarunvadde: Done. Much better now.

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
lotharking: You're right, but in this case, does this really need to be part of the SDK? I think this logic belongs in the controller instead.

tarunvadde: I think you are right. Moved it out of SDK. Please re-review.

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
```diff
@@ -0,0 +1,303 @@
...
+  input: AddServiceEndpointInput,
+): { services: DidDocumentService[]; added: ServiceEndpoint } {
+  const reserved = reservedErrorCode(input.type)
+  if (reserved) {
+    throw new ServiceEndpointError(
+      reserved,
+      `Service type '${input.type}' is managed automatically by the agent`,
+    )
+  }
+  if (!isValidServiceEndpoint(input.serviceEndpoint)) {
```
lotharking, line 125: I'm not sure about this in this context, but could you please review it? The service endpoint can be an array;are we taking that into account?

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
```diff
@@ -0,0 +1,303 @@
...
+      ServiceEndpointErrorCode.InvalidServiceEndpoint,
+      'serviceEndpoint does not conform to the DID-CORE shape (URI string, object, or array thereof)',
+    )
+  }
+
+  const existingIds = new Set(services.map(s => s.id))
+  const id = input.id
+    ? fullServiceId(agentDid, input.id)
+    : generateServiceId(agentDid, existingIds, input.type)
+  if (existingIds.has(id)) {
```
lotharking, line 136: This is a good practice, but what happens if someone wants to add another DIDComm v1 endpoint, or something similar that doesn't align with this requirement? Would it make more sense to use a blacklist instead?

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
tarunvadde: Yes, it's handled. isValidServiceEndpoint has an explicit array branch.

apps/vs-agent/src/controllers/admin/service-endpoints/ServiceEndpointsService.ts
tarunvadde: It's already a block-list, not an allow-list: the only reserved types are the agent-managed ones (DIDCommMessaging v2, LinkedVerifiablePresentation, VsAgentAdminAPI, plus internal AnonCredsRegistry/relativeRef). Everything else is allowed, and DIDComm v1 (did-communication) is intentionally permitted. It is only is only enforcing id uniqueness, so you can add multiple endpoints as long as the ids differ.

lotharking (approved):

PR #478, opened 2026-06-26
feat(vs-agent): publish VsAgentAdminAPI DID Document service entry

tarunvadde:
Lets other apps find the agent's admin API straight from its DID. The agent adds a `VsAgentAdminAPI` entry to its DID Document holding the admin API's public URL, so callers like the Verana MCP server or a browser can read the URL from the DID instead of being told it some other way. This is #466 (VSA-VTI-DIDDOC).

## How it works
- The entry is added only when the external admin API is on (`ADMIN_API_AUTH_MODE` includes `corporation`). Its value is `ADMIN_API_PUBLIC_URL`, copied exactly as given.
- It is re-added every time the agent publishes its DID Document, and removed if you turn the feature off.
- You cannot add, change, or delete it through the service-endpoint admin API; the agent manages it. (The block for that lives in #468.)

## Config
- Two new env vars: `ADMIN_API_PUBLIC_URL` and `ADMIN_API_AUTH_MODE` (the modes are `internal` and `corporation`).
- The agent refuses to start if they don't match: `corporation` set without a URL, or a URL set without `corporation`. The wider auth-mode work stays in #472.

## Tests
- Unit tests for the helper that builds the entry.

Closes #466.


packages/agent-sdk/src/did/adminApiService.ts
```diff
@@ -0,0 +1,16 @@
+import { DidDocument, DidDocumentService } from '@credo-ts/core'
+
+const ADMIN_API_SERVICE_TYPE = 'VsAgentAdminAPI'
```
lotharking, line 3: Probably this should be better used in a constant file, or maybe only use ´VsAgentAdminAPI´ directly.

apps/vs-agent/src/main.ts
```diff
@@ -157,6 +159,20 @@ const run = async () => {
...
 
+  if (ADMIN_API_PUBLIC_URL && !ADMIN_API_AUTH_MODE.includes('corporation')) {
+    serverLogger.error(
+      'ADMIN_API_PUBLIC_URL must not be set unless ADMIN_API_AUTH_MODE includes "corporation"',
+    )
+    process.exit(1)
+  }
+  if (ADMIN_API_AUTH_MODE.includes('corporation') && !ADMIN_API_PUBLIC_URL) {
+    serverLogger.error('ADMIN_API_PUBLIC_URL is required when ADMIN_API_AUTH_MODE includes "corporation"')
+    process.exit(1)
```
lotharking, line 198: > Auto-publish a VsAgentAdminAPI service entry in the agent's DID Document whenever the external listener is enabled (ADMIN_API_AUTH_MODE includes corporation)

Why is this considered mandatory? My understanding from the issue and its comments is that this is only required when `ADMIN_API_AUTH_MODE` includes `corporation`. If the external listener is not enabled, the `VsAgentAdminAPI` service should not be required.

packages/agent-sdk/src/did/adminApiService.ts
tarunvadde: It's a local const on purpose, so this PR builds on main without depending on #468, which defines the same VsAgentAdminAPI value for its guard. Once both merge I'll move it into one shared constant. I can inline it here for now if you'd rather not keep the const.

apps/vs-agent/src/main.ts
tarunvadde: The spec requires `ADMIN_API_PUBLIC_URL` when `ADMIN_API_AUTH_MODE` includes `corporation`, and forbids it otherwise. The check only fires on those bad setups: `corporation` on with no URL (the admin API would be enabled with no address to publish), or a URL set with `corporation` off. It just stops a broken start early, the normal case is not affected. If you'd rather #472 own all the auth-mode env validation, I can move it there.

apps/vs-agent/src/config/constants.ts
```diff
@@ -28,6 +28,12 @@ export const AGENT_PUBLIC_DID = process.env.AGENT_PUBLIC_DID
 export const PUBLIC_API_BASE_URL = process.env.PUBLIC_API_BASE_URL
 export const DEFAULT_PUBLIC_API_BASE_URL = 'http://localhost:3001'
 
+export const ADMIN_API_PUBLIC_URL = process.env.ADMIN_API_PUBLIC_URL
+export const ADMIN_API_AUTH_MODE = (process.env.ADMIN_API_AUTH_MODE ?? '')
```
lotharking:
```suggestion
export const ADMIN_API_AUTH_MODE = (process.env.ADMIN_API_AUTH_MODE ?? 'internal')
```

lotharking (approved):
I prefer to set up initial values.

PR #479, opened 2026-06-26
feat: check verifiable service requirements prior to accept vt flows

### LINKED ISSUE #428: Check Verifiable Service requirements prior to accept VT flows

lotharking:
**Changes**
- Implement DID rotation to did:peer (DIDComm v2) using DidRotateV2Service, including a temporary applicant-side rotation helper.
- Replace the Connections handshake with receiveImplicitInvitation (DIDComm v2).
- VS-CONN-VS trust check now resolves the counterparty's public DID via @verana-labs/verre, wired automatically from verifiablePublicRegistries; persist it as peerPublicDid on VtFlowRecord.
- Update integration tests to cover the DID rotation flows.


packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
```diff
@@ -0,0 +1,43 @@
...
+  deps: AssertVerifiableServiceDeps,
+): VtFlowAssertVerifiableServiceHook {
+  let verre: Promise<typeof import('@verana-labs/verre')> | undefined
+  let cache: ResolverConfig['cache']
+  return async ({ agentContext, peerDid }) => {
+    const logger = deps.logger ?? agentContext.config.logger
+    try {
+      const { resolveDID, InMemoryCache } = await (verre ??= importVerre('@verana-labs/verre'))
+      cache ??= new InMemoryCache()
+      const { verified, outcome, metadata } = await resolveDID(peerDid, {
```
tarunvadde, line 21: This returns `verified` and drops `outcome`. Verified against verre: `processDidDocument` returns `verified: true` for any structurally-valid service credential, with outcome defaulting to `NOT_TRUSTED` `outcome` only becomes `VERIFIED` when `refUrl.startsWith(registry.id) `and the registry is `production`. Since the gate ignores `outcome`, a peer that self-hosts a structurally-valid credential pointing its schema/registry at its own server passes VS-CONN-VS without resolving to a trusted Verana registry. The gate must require `outcome === VERIFIED` (and opt into `VERIFIED_TEST` only for test deployments), not just `verified`.

apps/vs-agent/src/utils/setupAgent.ts
```diff
@@ -88,10 +92,12 @@ export const setupAgent = async ({
         publicApiBaseUrl,
         endpoints,
         didcommVersions,
+        verifiablePublicRegistries: VERANA_INDEXER_BASE_URL
+          ? [{ id: 'verana', baseUrls: [VERANA_INDEXER_BASE_URL], production: false }]
```
tarunvadde: The registry is wired `id: 'verana'`, but real refs look like `vpr:verana:...`, and `verre` matches via `refUrl.startsWith(registry.id)`. `'vpr:verana:...'.startsWith('verana')` is false, so no real ref ever matches and `outcome` is always `NOT_TRUSTED` even for legitimate peers; production: false would also cap it at `VERIFIED_TEST`. So once the gate is fixed to require `VERIFIED`, this wiring rejects everyone. The `id` needs to be the actual ref prefix (e.g. `vpr:verana:`) and production set per environment.

apps/vs-agent/src/utils/setupAgent.ts
```diff
@@ -88,10 +92,12 @@ export const setupAgent = async ({
         publicApiBaseUrl,
         endpoints,
         didcommVersions,
+        verifiablePublicRegistries: VERANA_INDEXER_BASE_URL
```
tarunvadde: If `VERANA_INDEXER_BASE_URL` is unset, `verifiablePublicRegistries` is `undefined`, no `assertVerifiableService` hook is called, and `VtFlowService.assertVerifiableService` warns and permits. So one missing env var silently disables the whole VS-CONN-VS gate and accepts every peer. Can we require the indexer URL when vt-flow is enabled (refuse to start otherwise), or make the unconfigured-hook path reject.

apps/vs-agent/tests/vtFlow.test.ts
```diff
@@ -166,25 +188,175 @@ describe('vt-flow: two-agent integration', () => {
...
-  const validatorOutOfBand = await validatorAgent.didcomm.oob.createInvitation({
-    handshakeProtocols: [DidCommHandshakeProtocol.Connections],
+/**
+ * VS-CONN-VS trust gate: when an `assertVerifiableService` hook is configured, vt-flow runs trust
+ * resolution on the peer's public DID before acting. `resolveDID` (delegated to `@verana-labs/verre`
+ * in production) is mocked to always approve, so this only checks that a verified peer is let through
+ * and that both sides resolve the right DID: the Applicant the Validator (VS-SVC-4) on send, and the
+ * Validator the Applicant (VS-CONN-VS) on receive.
+ */
+describe('vt-flow: VS-CONN-VS trust gate (verre approves)', () => {
```
tarunvadde: Can we also add a negative case: hook returns `false` (and/or throws) and assert `sendOnboardingRequest/sendIssuanceRequest` rejects with `vt-flow.not-a-verifiable-service`, and the Validator does not reach `VALIDATING` on an OR/IR from an unverified peer

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -273,6 +275,19 @@ export class VtFlowOrchestrator {
     return match?.[1]?.toLowerCase()
   }
 
+  // TODO: temporary applicant-side rotation; remove once credo-ts `connections.rotate()` supports v2.
+  // Call after the first message has gone out on the public DID. See:
+  // https://github.com/verana-labs/vs-agent/pull/440/#issuecomment-4793219808
+  async rotateRequesterDidToPeer(connectionId: string): Promise<void> {
```
tarunvadde, line 273: `rotateRequesterDidToPeer` is only ever called from the test so the applicant connects with its public DID (via ourDid) and never rotates. Is this intended?

tarunvadde (changes requested):

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+export interface AssertVerifiableServiceDeps {
+  verifiablePublicRegistries: NonNullable<ResolverConfig['verifiablePublicRegistries']>
+  logger?: BaseLogger
+}
+
+// TODO: Remove this workaround after migrating to ESM
+const importVerre = new Function('s', 'return import(s)') as (
+  s: string,
+) => Promise<typeof import('@verana-labs/verre')>
```
genaris: I think we are on time to fix this on verre's packaging so we don't have to add this workaround. I raised an issue https://github.com/verana-labs/verre/issues/111 for this.

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
```diff
@@ -0,0 +1,43 @@
...
+
+/**
+ * VS-CONN-VS gate: delegates trust resolution to `@verana-labs/verre` (`resolveDID`), which verifies
+ * the peer's VTC signatures, schema bindings, digestSRI, issuer permissions and recurses to the
+ * Ecosystem trust root. We only map the result to a boolean (fail-closed on error, TR-8).
+ */
```
genaris: Probably too verbose commenting here, as usual

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
```diff
@@ -0,0 +1,43 @@
...
+export function assertVerifiableService(
+  deps: AssertVerifiableServiceDeps,
```
genaris: this "deps"/*Deps interface doesn't seem consistent with the coding styling we usually follow. Usually we name these method parameters as "options"

packages/vt-flow/src/repository/VtFlowRecord.ts
```diff
@@ -45,6 +45,8 @@ export interface VtFlowStorageProps {
 
   errorMessage?: string
 
+  peerPublicDid?: string
```
genaris: What do you mean by "peerPublicDid" and why is it needed. I guess you are referring to the "other party's public DID", which you can use to perform trust resolution.

If that's the case, first of all I'd say that prefixing it with "peer" is not a good idea, since we associate peer with the did:peer method (i.e. the one we are rotating from the public DID and using for the further communication).

Second, I think there it might not be a specific need to store the DID on the VtFlowRecord, since it is already on the DidCommConnectionRecord (since we are always using implicit invitations, it will be stored as `invitationDid`). Of course, guards and checks are always needed to make sure the invitation DID is a public one (i.e. it exists and it is not a did:peer).

packages/vt-flow/src/services/VtFlowService.ts
```diff
@@ -147,7 +151,9 @@ export class VtFlowService {
   ): Promise<VtFlowRecord> {
     const { message, agentContext } = messageContext
     const connection = messageContext.assertReadyConnection()
-    await this.assertVerifiableService(agentContext, connection.id)
+    const peerDid = connection.theirDid
+    if (!peerDid) throw new CredoError(`vt-flow: ready connection '${connection.id}' has no theirDid`)
+    await this.assertVerifiableService(agentContext, peerDid, connection.id)
```
genaris: Probably this could be simplified to 

```
this.assertVerifiableService(agentContext, connection.id)
```

or 
 
```
this.assertVerifiableService(agentContext, connection.invitationDid)
```

If you want to do the pure trust resolution (which is what I see being done right now, i.e. no use for connection.id at all).


BTW I think that using the very same name for both the internal wrapper method and the hook is very inconvenient. Can you use something different like 'checkIsVerifiableService' or something that can make us distinguish from the hook? Or maybe suffix the hook, like`assertVerifiableServiceCallback`?

genaris (changes requested):

packages/agent-sdk/src/plugins/setupBaseDidComm.ts
```diff
@@ -30,13 +30,16 @@ import { defaultDocumentLoader } from '../did/CachedDocumentLoader'
...
+import { assertVerifiableService, type AssertVerifiableServiceDeps } from '../vtFlow'
 
 export interface BaseDidCommPluginOptions {
   walletConfig: AskarModuleConfigStoreOptions
   publicApiBaseUrl: string
   endpoints: string[]
   vtFlow?: VtFlowModuleConfigOptions
   didcommVersions?: DidCommVersion[]
+  /** When set, wires the default VS-CONN-VS trust-resolution hook (verre) into vt-flow unless `vtFlow.assertVerifiableService` is already provided. */
+  verifiablePublicRegistries?: AssertVerifiableServiceDeps['verifiablePublicRegistries']
```
genaris: 🤯

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -273,6 +275,19 @@ export class VtFlowOrchestrator {
...
+    const agentContext = this.agent.context
+    const didRotateV2 = agentContext.dependencyManager.resolve(DidCommDidRotateV2Service)
+    const routing = await agentContext.dependencyManager
+      .resolve(DidCommRoutingService)
+      .getRouting(agentContext, {})
+    const connection = await this.agent.didcomm.connections.getById(connectionId)
+    await didRotateV2.rotateOurDid(agentContext, connection, routing)
+  }
+
```
genaris: There is a new Credo alpha supporting DID Rotation over V2 connections (see https://github.com/openwallet-foundation/credo-ts/pull/2845). The new version is v0.7.1-pr-2704-20260630143332

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
lotharking: I'm not sure how we should handle this. At the moment, a `VERIFIED_TEST` is considered valid (e.g., `chatbot demo dev`). How do you recommend handling this, @genaris?

genaris: I think that's a good question. We can discuss about it in this week's WG call.

apps/vs-agent/src/utils/setupAgent.ts
lotharking: Use a warning check to handle this. By the way, I left a TODO to make this feature mandatory in the future

patches/@credo-ts__webvh@0.7.1-pr-2704-20260630143332.patch
genaris, line 1: @tarunvadde do you remember if this patch was still needed? You analyzed that recently but my short-term memory is working pretty badly nowadays.

genaris (approved):

patches/@credo-ts__webvh@0.7.1-pr-2704-20260630143332.patch
tarunvadde: No, it wasn't needed. We can remove it safely.

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
tarunvadde: This is what I was suggesting. The gate accepts the peer no matter what the trust result is. It returns `verified` and never checks `outcome`.

Quick correction on VERIFIED_TEST, since it changes how we frame this. VERIFIED_TEST is not what's being accepted. With the current config (`id: 'verana'`), the demo's schema refs don't match the registry, so on this branch they come back as NOT_TRUSTED, not VERIFIED_TEST. They pass only because the gate ignores `outcome` and lets NOT_TRUSTED through. So VERIFIED_TEST isn't in play yet. Worth checking on the live demo, but that's what the code here does.

Two problems, and they need to be fixed together. Fixing only the gate would reject every peer.

**Problem 1: the gate ignores the trust outcome.** `verified` only means the peer is a verifiable service. verre returns `verified` and `outcome` as two separate fields (in `didValidator.js` returns `{ verified: true, outcome }`). So a NOT_TRUSTED peer is still accepted today.

Make it an allowlist. Accept VERIFIED and VERIFIED_TEST, reject everything else:

```ts
export interface AssertVerifiableServiceDeps {
  verifiablePublicRegistries: NonNullable<ResolverConfig['verifiablePublicRegistries']>
  allowTestTrust?: boolean // true while the registry is non-production
  logger?: BaseLogger
}

// inside the returned hook:
const { resolveDID, InMemoryCache, TrustResolutionOutcome } =
  await (verre ??= importVerre('@verana-labs/verre'))
cache ??= new InMemoryCache()
const { verified, outcome, metadata } = await resolveDID(peerDid, {
  verifiablePublicRegistries: deps.verifiablePublicRegistries,
  cache,
})
const trusted =
  verified &&
  (outcome === TrustResolutionOutcome.VERIFIED ||
    (deps.allowTestTrust === true && outcome === TrustResolutionOutcome.VERIFIED_TEST))
if (!trusted) {
  logger.warn(`[vt-flow] VS-CONN-VS rejected '${peerDid}': verified=${verified} outcome=${outcome} ${metadata?.errorMessage ?? ''}`)
}
return trusted
```

`TrustResolutionOutcome` is a runtime enum, so take it from the dynamic `importVerre` (not a static import, because of the ESM workaround). `allowTestTrust` comes from the same config as the registry `production` flag: true while `production` is false (test creds pass), false in production (only VERIFIED passes).

**Problem 2: the gate fix alone rejects everyone, because the registry `id` is wrong.** Today it's `{ id: 'verana' }` (setupAgent.ts). verre matches with `refUrl.startsWith(registry.id)`. The real refs don't start with "verana":
- on-chain: `vpr:verana:<chainId>/cs/v1/js/<id>` (stateMutations.ts:231)
- self-TR / demo: `https://<agent>/vt/...` (setupSelfTr.ts)

Both come back NOT_TRUSTED. That's why fixing only the gate rejects everyone.

Set `id` to the real prefix so the ref matches and points to a real URL:

```ts
// setupAgent.ts  (was id: 'verana')
verifiablePublicRegistries: VERANA_INDEXER_BASE_URL
  ? [{ id: `vpr:verana:${VERANA_CHAIN_ID}`, baseUrls: [VERANA_INDEXER_BASE_URL], production: false }]
  : undefined
// and pass allowTestTrust (= !production) through to assertVerifiableService
```

Now `vpr:verana:<chainId>/cs/v1/js/<id>` resolves to verified-test, and verre rewrites the schema URL to `${VERANA_INDEXER_BASE_URL}/cs/v1/js/<id>`. Two things to line up:
1. the indexer needs to serve that path (`<base>/cs/v1/js/<id>`, returning the raw JSON schema). @lotharking can you confirm the base and path?
2. the self-TR demo uses https refs, which still won't match. So the demo moves to on-chain (`vpr:verana:`) creds, or gets its own registry entry. We should decide what the demo presents.

I checked this against verre 0.3.1's `resolveTrustRegistry`: with `id: 'verana'` both refs are NOT_TRUSTED. With `id: 'vpr:verana:<chainId>'` the on-chain ref comes back verified-test, schemaUrl `<indexer>/cs/v1/js/<id>`.

My apologies for this long reply and I got this clear version from claude but this covers what I was implying in my earlier comment.

tarunvadde (approved):
Just added one big comment. If possible, can you take it up and verify in the same PR or we can discuss further on this tomorrow and I will approve the PR for now.

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
lotharking: Hey Tarum, thinking about it a bit more: the second point is partially solved; I included the /verana/ path in the codebase so the library resolves the URLs correctly. On another note, the verre dependency was updated to a new version with proper CJS support, so that's covered too. The only point still inconclusive is the one about VERIFIED_TEST, and I'm not fully sure about it yet because we first need to define how the old demo is going to be maintained.

tarunvadde: Good thing that you created a separate issue for that 👍

PR #480, opened 2026-06-29
feat(vs-agent): add revocation-registry kebab-case route alias

tarunvadde:
Closes #467.

Adds the kebab-case alias `/v1/credential-types/revocation-registry` (GET/POST/DELETE) next to the existing camelCase `/revocationRegistry`. CamelCase stays for backward compat.

The issue's other three items were already spec-compliant, so no code change: `GET /v1/health` returns 200, `GET /v1/qr` takes the spec query params (`size`, `padding`, `level`, `bcolor`, `fcolor`, `legacy`).


genaris (approved):
Is there an elegant way to mark the older route as "deprecated" in the API docs? Or at least mention that as as note on it?

PR #482, opened 2026-06-29
feat(agent-sdk): harden indexer subscription

tarunvadde:
Implements #474. Hardens the indexer subscription and adds a corp-scoped option.

Hardening (`IndexerWebSocketService`):
- Skips duplicates by `(tx_hash, message_index)`, not just `block_height`.
- Processes events one at a time, in `(tx_index, message_index)` order.
- On reconnect, a failed block no longer moves the watermark forward or gets applied twice.
- Moves the last-seen-block watermark only after a handler succeeds.
- Pages through the whole REST catch-up.
- Keeps the bootstrap order (catch up before accepting incoming DIDComm) and logs the zero-balance / no-authorization warning.

Corp-scoped subscription:
- Adds the corp-scoped subscribe (`{ action: "subscribe", corporationId }`) and sends `?corporation_id=` on the catch-up. DID scope stays the default.
- Turned on with `VERANA_INDEXER_SUBSCRIPTION_SCOPE=corporation`, using `VERANA_CORPORATION_ID`.

Closes #474.

PR #484, opened 2026-06-30
feat(vs-agent): add and validate v4 config and env surface

tarunvadde:
Adds the v4 container env vars and validates them at startup. If anything required is missing or malformed, the agent fails fast with a clear message instead of starting in a broken state.

New vars:
- `VERANA_CORPORATION_ID` (required)
- `AGENT_MODE` (`standalone` | `delegated`, default `standalone`)
- `AGENT_DELEGATED_PARENT_VS_DID` (required when `AGENT_MODE=delegated`)
- `ADMIN_API_CORPORATION_ALLOWED_ACCOUNTS` (optional allowlist, parsed only. the allowlist is applied by the auth layer in #472)

Now required at startup: `VERANA_CORPORATION_ID`, `VERANA_RPC_ENDPOINT_URL`, `VERANA_INDEXER_BASE_URL`, `VERANA_ACCOUNT_MNEMONIC`.

- This is going to be a behaviour change because the agent no longer starts without a corporation id, RPC, indexer, and mnemonic. This matches the v4 spec, which requires these at startup.

Notes:
- `ADMIN_API_AUTH_MODE` and `ADMIN_API_PUBLIC_URL` and their validation are delivered in #478.
- `setupAgent.ts` is named in the issue but needs no change: it reads none of these vars. main.ts validates them and passes the chain service into `setupAgent` as a parameter.

Part of #464.


apps/vs-agent/src/config/constants.ts
```diff
@@ -126,6 +126,18 @@ export const VERANA_INDEXER_BASE_URL = process.env.VERANA_INDEXER_BASE_URL
 export const VERANA_ACCOUNT_MNEMONIC = process.env.VERANA_ACCOUNT_MNEMONIC
 export const VERANA_RPC_ENDPOINT_URL = process.env.VERANA_RPC_ENDPOINT_URL
 export const VERANA_CHAIN_ID = process.env.VERANA_CHAIN_ID
+export const VERANA_CORPORATION_ID = process.env.VERANA_CORPORATION_ID
+
+// Agent mode: standalone (default) or delegated
```
lotharking: Is this correctly documented?
```suggestion
```

tarunvadde: Yes but removed as per your suggestion

apps/vs-agent/src/main.ts
```diff
@@ -150,6 +153,22 @@ const run = async () => {
...
+  }
+  if (!['standalone', 'delegated'].includes(AGENT_MODE)) {
+    configErrors.push(`AGENT_MODE must be 'standalone' or 'delegated' (got '${AGENT_MODE}')`)
+  }
+  if (AGENT_MODE === 'delegated' && !AGENT_DELEGATED_PARENT_VS_DID) {
+    configErrors.push('AGENT_DELEGATED_PARENT_VS_DID is required when AGENT_MODE=delegated')
+  }
+  if (configErrors.length > 0) {
+    serverLogger.error(`Invalid configuration:\n- ${configErrors.join('\n- ')}`)
+    process.exit(1)
```
lotharking, line 163: This needs to be removed, right?

tarunvadde: This is the optional version now. I removed the hard-required checks, so it does not break demos. If nothing is set, `AGENT_MODE` defaults to `standalone`, all checks pass, and the agent starts fine. It only exits on config that is explicitly wrong, like a `VERANA_CORPORATION_ID` that is not a number.

`AGENT_MODE` and `AGENT_DELEGATED_PARENT_VS_DID` are part of the v4 config from #464. They are added ahead of their consumer, the same as `VERANA_CORPORATION_ID` which #482 uses.  I'd lean to keeping them since they're from #464, but can remove if you prefer.

lotharking: Is this condition mandatory?
```ts
if (VERANA_CORPORATION_ID && !/^\d+$/.test(VERANA_CORPORATION_ID)) {
  configErrors.push('VERANA_CORPORATION_ID must be a non-negative integer')
}
```

tarunvadde: No, it is not mandatory. It is a guard: `VERANA_CORPORATION_ID` is used by #482, which does Number(`VERANA_CORPORATION_ID`). If it is set to something that is not a number, that becomes `NaN` and the indexer rejects the corporation query. This check turns that into a clear error at startup instead. Happy to remove it if you prefer, since the only consumer is #482.

lotharking (approved):

PR #486, opened 2026-07-01
feat: Developer overridable handler extension points

lotharking:
This PR updates the documentation and exports `IndexerHandlerContext` for event handlers.
According to the issue, the requested functionality is already implemented, but I'm not sure if these changes are sufficient. I'd appreciate your feedback.


tarunvadde (approved):
LGTM, I think this is sufficient.

packages/agent-sdk/src/blockchain/handlers/IndexerHandlerRegistry.ts
```diff
@@ -1,7 +1,7 @@
 import { VsAgent } from '../../agent'
 import { IndexerActivity, VeranaSyncState } from '../types'
 
-interface IndexerHandlerContext {
+export interface IndexerHandlerContext {
   agent: VsAgent
   block_height: number
```
genaris: Why this is snake_case and other fields camelCase?

genaris (commented):
This PR seems OK to me, but since I think you are tryign to close #475 , I want to give my 2 cents (also called "mouthpieces"):

The spec describes vs agent as a whole, in a high-level basis, so when it says:

> Each notification must be associated with a specific handler interface in the VS Agent. A default implementation will be provided to handle the most important notifications. Developers can implement their own handlers to override VS Agent default handlers (or provide an implementation for notifications not handled by the default implementation).

It is not talking about an SDK or a particular component, but rather the container (in the very beginning, it states "VS Agent is a container that provides the full stack required to operate a Verifiable Service. It bundles, in a single deployable unit"). So I don't think this mechanism provided by the SDK will be enough to satisfy the aforementioned statement.

Currently, vs-agent app constructs `IndexerWebSocketService` with no handlerRegistry, so the stock container always uses the defaults. There is currently no way for a backend behind the container to hook into indexer notifications: webhookEvent.ts only forwards DIDComm domain events (connection/message/presentation), never on-chain notifications.

So if we want to close #475, we'll need to do something like (please just take it as an idea/draft): 

1. Forward notifications as webhook events. Add an indexer-notification webhook (reuse EVENTS_BASE_URL, e.g. POST `${EVENTS_BASE_URL}/indexer-notification` with the IndexerActivity + block/tx context). This alone satisfies "provide an implementation for notifications not handled by the default implementation" — the backend can react to any msg, including the many that are currently N/A in the spec tables.

2. Per-msg override config to satisfy "override VS Agent default handlers." An env var like `INDEXER_DEFAULT_HANDLERS_DISABLED=CreateNewCredentialSchema,RevokeParticipant` (or *).

apps/vs-agent/src/config/constants.ts
```diff
@@ -132,6 +132,12 @@ export const VERANA_INDEXER_BASE_URL = process.env.VERANA_INDEXER_BASE_URL
 export const VERANA_ACCOUNT_MNEMONIC = process.env.VERANA_ACCOUNT_MNEMONIC
 export const VERANA_RPC_ENDPOINT_URL = process.env.VERANA_RPC_ENDPOINT_URL
 export const VERANA_CHAIN_ID = process.env.VERANA_CHAIN_ID
+export const INDEXER_DEFAULT_HANDLERS_DISABLED: string[] = (
```
genaris:
```suggestion
export const VERANA_INDEXER_DEFAULT_HANDLERS_OVERRIDE: string[] = (
```

Just to make sure what our intentions are. And Prefixing to "VERANA" as other related features.

genaris (approved):

PR #490, opened 2026-07-03
feat: applicant credential verification and acceptance

lotharking:
**Changes**
- Migrated VeranaIndexerService reads to the /v4 indexer routes and added getParticipantSession and getDigest.
- Added ParticipantRole/ParticipantState enums and ParticipantSessionDto/DigestDto; ParticipantDto now includes participant_state.
- Rewrote VtFlowOrchestrator.verifyOfferedCredential to be indexer-only
- Added onCredentialCompleted (publish LinkedVP + conditional TriggerResolver).
- Wired the verifyCredential and onCompleted vt-flow hooks in setupAgent.
> Note: requires https://github.com/verana-labs/verana-indexer/pull/308


apps/vs-agent/src/utils/setupAgent.ts
```diff
@@ -101,6 +103,11 @@ export const setupAgent = async ({
         ]
       : undefined
 
+  const indexer = VERANA_INDEXER_BASE_URL
+    ? new VeranaIndexerService({ baseUrl: VERANA_INDEXER_BASE_URL, logger })
+    : undefined
+  const orchestratorRef: { current?: VtFlowOrchestrator } = {}
```
genaris: I don't see why can't we just do `let orchestrator: VtFlowOrchestrator | undefined`. This pattern of wrapping object reference seems similar to useRef hook in React but I don't think it is the best solution in backend.

Is it something related to the linter?

packages/agent-sdk/src/blockchain/VeranaChainService.ts
```diff
@@ -313,17 +328,27 @@ export class VeranaChainService {
...
-    const result = await this.broadcastMsg(veranaTypeUrls.MsgTriggerResolver, value)
+    const result = await this.broadcastMsg(veranaTypeUrls.MsgTriggerResolver, value, true)
     return { txHash: result.transactionHash }
   }
 
-  private async broadcastMsg(typeUrl: string, value: object): Promise<DeliverTxResponse> {
+  private async broadcastMsg(
+    typeUrl: string,
+    value: object,
+    useSessionSigner = false,
```
genaris: I think wrapping in an options object would be nice.



```suggestion
options: {
    typeUrl: string,
    value: object,
    useSessionSigner = false,
  }
```

for better readability when using it (e.g. this.broadcastMsg({ typeUrl: veranaTypeUrls.MsgCreateOrUpdateParticipantSession, value, useSessionSigner: true })

apps/vs-agent/src/utils/setupAgent.ts
lotharking: Yes, the linter breaks due to never reassigned, but in this case we reassigned `current`, that is an internal property.

genaris: I did the changes locally and didn't face that problem. orchestrator is assigned in line 179, so that error shouldn't be there.

Are you sure you assigned it as

```
let orcherstrator: VtFlowOrchestrator | undefined
```

?

lotharking: The problem is not the logic, it is related to the linter rules. In this case, it is possible to use `// eslint-disable-next-line prefer-const` to disable the rule validation. This has been implemented to guarantee more consistency with the source code.

genaris (approved):

PR #493, opened 2026-07-06
feat(agent-sdk): complete indexer notification default handlers

tarunvadde:
Closes #471

- `RenewParticipantOP` progresses the applicant flow (reconnect and resend, as per the spec's Reconnection rule).
- `RevokeParticipant` / `SlashParticipantTrustDeposit`: a HOLDER's linked VP and stored credential are removed, and flows are terminated with an ERROR problem report (`PARTICIPANT_REVOKED` / `PARTICIPANT_SLASHED`).
- New handlers: corporation events (with a DID-rotation warning), `ArchiveEcosystem`, `CreateRootParticipant`, `SelfCreateParticipant`, `TriggerResolver`.
- Handler keys verified against a live verana-indexer:dev, where the spec strings do not match, the code follows the indexer.

Note: the two GF handlers and `TriggerResolver` stay inactive until the indexer gaps are fixed (issue filed [here](https://github.com/verana-labs/verana-indexer/issues/310)).


packages/agent-sdk/src/blockchain/handlers/defaultHandlers.ts
```diff
@@ -85,8 +86,9 @@ export const defaultHandlers: IndexerEventHandler[] = [
     msg: 'RenewParticipantOP',
     handle: async (activity, ctx) => {
       ctx.agent.config.logger.info(
-        `[IndexerWS] RenewParticipantOP entity=${activity.entity_id} block=${ctx.blockHeight} — TODO §5.1: progress credential acquisition flow (applicant renewal)`,
+        `[IndexerWS] RenewParticipantOP entity=${activity.entity_id} block=${ctx.blockHeight}`,
       )
+      await startParticipantOPAutoFlow(ctx.agent, activity)
```
lotharking, line 91: Please review this. We shouldn't create a duplicate flow; the specification uses the reconnect flow for this.

lotharking (commented):
> Reconcile event_type strings with the indexer's actual Cosmos action names. The spec uses SetParticipantOPtoValidated and IncreaseActiveGovernanceFrameworkVersion; the code currently uses SetParticipantOPToValidated and IncreaseActiveGFVersion. Verify against real indexer output and align — handlers will silently not fire if the strings mismatch

What about this? The gap has already been resolved. Please try again and let me know if it's still an issue.

packages/agent-sdk/src/blockchain/handlers/defaultHandlers.ts
tarunvadde: Thanks for pointing this out. Fixed, now the flow will be on renewal the applicant resends the OR with the same `participant_session_id` (reusing the connection if still open), and the validator re-attaches the finished flow and re-runs it instead of creating a new one. Also added a peer check on re-attach since session ids are public on-chain. 

One important note is that the spec recommends a fresh session id with matching by `participant_id`, but validator matching is by session id today, so reusing it is what interoperates. Can switch once `participant_id` matching exists.

lotharking (approved):

PR #495, opened 2026-07-08
feat(agent-sdk): authorization notifications and cache

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/473

- New `AuthorizationService`: caches the agent's VSOA records per participant and answers `canSign` / `hasFeegrant` synchronously. Operator authorization checks (agent or caller account) query the chain on demand, since there is no indexer event to cache them against. This is the read surface https://github.com/verana-labs/vs-agent/issues/472 will consume. The `AUTHORIZATION_SERVICE` Nest provider is already wired.
- `registerAuthorizationHandlers` layers the cache refresh on top of the existing pp handlers (grants and expiration updates refresh, revoke/slash/cancel invalidate then refresh) and registers the six spec Authorization Notification handlers (MOD-DE-MSG-1..6). Those six stay dormant until the indexer emits delegation events (issue to be filed on verana-indexer).
- The cache is seeded at startup and retried until it succeeds, so it cannot stay silently empty.
- Also re-removes the dual-account session signing that the https://github.com/verana-labs/vs-agent/pull/490 merge brought back after https://github.com/verana-labs/vs-agent/pull/492 removed it: one `vs_operator` signs everything since node `v0.10.1-dev.21`.
- The e2e drives the cache through the real path (chain tx -> indexer WS event -> handler -> cache) across grant, activate and revoke, plus the caller checks. Lapsed grants with a period count as active because the chain auto-renews them at check time.

**Note:** `feegrant` state is binary (`with_feegrant` present or absent), matching the spec. The issue's "remaining feegrant validity" recompute has no spec row and was left out on purpose.

Another note on comments: I kept a few one-liners on purpose where the code would be misleading without them (the period auto-renew rule, the fail-closed blank account check, and the dormant delegation handlers). If any feel too much, say so and I will drop them.

### DIFF
--- a/apps/vs-agent/src/admin.module.ts
+++ b/apps/vs-agent/src/admin.module.ts
@@ -48,10 +48,6 @@ export class VsAgentModule {
         useFactory: () => agentRef.get(),
       },
       {
-        provide: 'AUTHORIZATION_SERVICE',
-        useFactory: () => agentRef.get().authorizationService,
-      },
-      {
         provide: 'PUBLIC_API_BASE_URL',
         useFactory: () => publicApiBaseUrl,
       },
--- a/apps/vs-agent/src/admin.module.ts
+++ b/apps/vs-agent/src/admin.module.ts
@@ -48,6 +48,10 @@ export class VsAgentModule {
         useFactory: () => agentRef.get(),
       },
       {
+        provide: 'AUTHORIZATION_SERVICE',
+        useFactory: () => agentRef.get().authorizationService,
+      },
+      {
         provide: 'PUBLIC_API_BASE_URL',
         useFactory: () => publicApiBaseUrl,
       },
--- a/apps/vs-agent/src/main.ts
+++ b/apps/vs-agent/src/main.ts
@@ -121,6 +121,8 @@ export const startServers = async (agent: VsAgent, serverConfig: ServerConfig) =
   return { httpServer, webSocketServer }
 }
 
+const AUTHORIZATION_SEED_RETRY_MS = 30_000
+
 const run = async () => {
   const serverLogger = new TsLogger(ADMIN_LOG_LEVEL, 'Server')
 
@@ -255,8 +257,7 @@ const run = async () => {
     })
     await veranaChain.start()
 
-    authorizationService = new AuthorizationService(veranaChain, serverLogger)
-    // The cache must not stay silently empty: retry the seed until it succeeds once.
+    authorizationService = new AuthorizationService({ chain: veranaChain, logger: serverLogger })
     const seedAuthorizationCache = async (): Promise<boolean> =>
       authorizationService!
         .refreshForOperator()
@@ -270,13 +271,16 @@ const run = async () => {
     if (!(await seedAuthorizationCache())) {
       const retry = setInterval(async () => {
         if (await seedAuthorizationCache()) clearInterval(retry)
-      }, 30_000)
+      }, AUTHORIZATION_SEED_RETRY_MS)
       retry.unref()
     }
 
     try {
       const balance = await veranaChain.getBalance()
-      if (authorizationService.listVsoaRecords().length === 0 && Number(balance.amount) === 0) {
+      if (
+        authorizationService.listVsOperatorAuthorizationRecords().length === 0 &&
+        Number(balance.amount) === 0
+      ) {
         serverLogger.warn(
           `[VeranaChain] Operator account ${veranaChain.address} has no VSOperatorAuthorization and zero ${balance.denom} balance; on-chain operations will fail until it is granted authorization or funded.`,
         )
--- a/apps/vs-agent/src/main.ts
+++ b/apps/vs-agent/src/main.ts
@@ -4,6 +4,7 @@ import { parseDid, utils } from '@credo-ts/core'
 import { NestFactory } from '@nestjs/core'
 import { KdfMethod } from '@openwallet-foundation/askar-nodejs'
 import {
+  AuthorizationService,
   HttpInboundTransport,
   setupSelfTr,
   VsAgent,
@@ -12,6 +13,7 @@ import {
   VeranaChainService,
   IndexerWebSocketService,
   buildDefaultIndexerHandlerRegistry,
+  registerAuthorizationHandlers,
 } from '@verana-labs/vs-agent-sdk'
 import * as express from 'express'
 import * as fs from 'fs'
@@ -242,6 +244,7 @@ const run = async () => {
 
   // Connect to Verana blockchain for on-chain transactions
   let veranaChain: VeranaChainService | undefined
+  let authorizationService: AuthorizationService | undefined
   if (VERANA_RPC_ENDPOINT_URL && VERANA_ACCOUNT_MNEMONIC) {
     veranaChain = new VeranaChainService({
       rpcUrl: VERANA_RPC_ENDPOINT_URL,
@@ -252,10 +255,28 @@ const run = async () => {
     })
     await veranaChain.start()
 
+    authorizationService = new AuthorizationService(veranaChain, serverLogger)
+    // The cache must not stay silently empty: retry the seed until it succeeds once.
+    const seedAuthorizationCache = async (): Promise<boolean> =>
+      authorizationService!
+        .refreshForOperator()
+        .then(() => true)
+        .catch(error => {
+          serverLogger.error(
+            `[Authorization] failed to seed the authorization cache: ${(error as Error).message}`,
+          )
+          return false
+        })
+    if (!(await seedAuthorizationCache())) {
+      const retry = setInterval(async () => {
+        if (await seedAuthorizationCache()) clearInterval(retry)
+      }, 30_000)
+      retry.unref()
+    }
+
     try {
       const balance = await veranaChain.getBalance()
-      const authorized = await veranaChain.hasVsOperatorAuthorization()
-      if (!authorized && Number(balance.amount) === 0) {
+      if (authorizationService.listVsoaRecords().length === 0 && Number(balance.amount) === 0) {
         serverLogger.warn(
           `[VeranaChain] Operator account ${veranaChain.address} has no VSOperatorAuthorization and zero ${balance.denom} balance; on-chain operations will fail until it is granted authorization or funded.`,
         )
@@ -299,6 +320,7 @@ const run = async () => {
     masterListCscaLocation: MASTER_LIST_CSCA_LOCATION,
     autoUpdateStorageOnStartup: AGENT_AUTO_UPDATE_STORAGE_ON_STARTUP,
     veranaChain,
+    authorizationService,
     adminApiServiceEndpoint,
   })
 
@@ -356,6 +378,7 @@ const run = async () => {
         `[IndexerWS] Default handlers disabled: ${VERANA_INDEXER_DEFAULT_HANDLERS_OVERRIDE.join(', ')}`,
       )
     }
+    if (authorizationService) registerAuthorizationHandlers(handlerRegistry, authorizationService)
 
     const indexerCorporationId =
       VERANA_INDEXER_SUBSCRIPTION_SCOPE === 'corporation' && VERANA_CORPORATION_ID
--- a/packages/agent-sdk/src/blockchain/AuthorizationService.ts
+++ b/packages/agent-sdk/src/blockchain/AuthorizationService.ts
@@ -5,9 +5,16 @@ import { CachedVsOperatorAuthorizationRecord, DurationParam } from './types'
 
 // A lapsed grant that carries a period is still valid: the chain rolls the expiration
 // forward on the next check (VPR spec AUTHZ-CHECK-3 step 4, AUTHZ-CHECK-1 step 2).
-function isActive(expiration?: Date, period?: DurationParam): boolean {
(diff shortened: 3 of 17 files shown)


apps/vs-agent/src/admin.module.ts
```diff
@@ -47,6 +47,10 @@ export class VsAgentModule {
         provide: 'VSAGENT',
         useFactory: () => agentRef.get(),
       },
+      {
+        provide: 'AUTHORIZATION_SERVICE',
+        useFactory: () => agentRef.get().authorizationService,
+      },
```
lotharking: Why include this provider?

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
```diff
@@ -0,0 +1,72 @@
...
+const VSOA_REVOKE_MSGS = [
+  'RevokeParticipant',
+  'SlashParticipantTrustDeposit',
+  'CancelParticipantOPLastRequest',
+] as const
+
+// The six Authorization Notifications from the spec (MOD-DE-MSG-1..6). The indexer does
+// not deliver delegation events yet (https://github.com/verana-labs/verana-indexer/issues/320),
+// so these stay dormant; the pp triggers above cover the VSOA lifecycle in the meantime.
+const DELEGATION_MSGS = [
```
lotharking: What about `UpdateVSOperatorAuthorizationExpiration`?

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  for (const msg of VSOA_GRANT_OR_UPDATE_MSGS) {
+    wrap(msg, () => authorizationService.refreshForOperator())
+  }
+
+  for (const msg of VSOA_REVOKE_MSGS) {
+    wrap(msg, async activity => {
+      const participantId = Number(activity.entity_id)
+      if (Number.isFinite(participantId)) authorizationService.invalidateParticipant(participantId)
+      await authorizationService.refreshForOperator()
```
lotharking, line 59: > invalidate the cached records; stop signing for the affected participants until re-granted

Remember, we need to invalidate the cache before refresh.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
```diff
@@ -0,0 +1,101 @@
+import { BaseLogger } from '@credo-ts/core'
+
+import { VeranaChainService } from './VeranaChainService'
+import { CachedVsoaRecord, DurationParam } from './types'
+
+// A lapsed grant that carries a period is still valid: the chain rolls the expiration
```
lotharking, line 6: Please review the comments. I think there are quite a few of them in this PR.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
```diff
@@ -0,0 +1,101 @@
...
+export class AuthorizationService {
+  private vsoaByParticipant = new Map<number, CachedVsoaRecord>()
+  private lastRefreshAt = 0
+  private readonly minRefreshIntervalMs: number
+
+  constructor(
+    private readonly chain: VeranaChainService,
+    private readonly logger: BaseLogger,
+    options: { minRefreshIntervalMs?: number } = {},
+  ) {
```
lotharking: I prefer a group in another object for consistency
```suggestion
  constructor(
    private readonly config: AuthorizationServiceConfig
  ) {
```

apps/vs-agent/src/main.ts
```diff
@@ -252,10 +255,28 @@ const run = async () => {
...
+        .catch(error => {
+          serverLogger.error(
+            `[Authorization] failed to seed the authorization cache: ${(error as Error).message}`,
+          )
+          return false
+        })
+    if (!(await seedAuthorizationCache())) {
+      const retry = setInterval(async () => {
+        if (await seedAuthorizationCache()) clearInterval(retry)
+      }, 30_000)
```
lotharking: I would prefer to replace this with a constant to provide more context for these values.

apps/vs-agent/src/main.ts
```diff
@@ -252,10 +255,28 @@ const run = async () => {
...
+      retry.unref()
+    }
+
     try {
       const balance = await veranaChain.getBalance()
-      const authorized = await veranaChain.hasVsOperatorAuthorization()
-      if (!authorized && Number(balance.amount) === 0) {
+      if (authorizationService.listVsoaRecords().length === 0 && Number(balance.amount) === 0) {
```
lotharking: What is Vsoa? I can't find it on the spec.

lotharking (changes requested):

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
tarunvadde: There's no such tx. MOD-DE-MSG-9 is a module call, only invoked from SetParticipantOPToValidated and SetParticipantEffectiveUntil, and both of those already trigger a refresh here. The de keeper does emit update_vs_operator_authorization though, which https://github.com/verana-labs/verana-indexer/pull/324/ forwards as UpdateVSOperatorAuthorization, so I added that event to the list.

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
tarunvadde: The pp revoke path already invalidates first. Also moved RevokeVSOperatorAuthorization into that same invalidate-then-refresh path, since https://github.com/verana-labs/verana-indexer/pull/324/ carries the participant_id as entity_id.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
tarunvadde: Trimmed.

apps/vs-agent/src/main.ts
tarunvadde: It's the spec's own shorthand: the VSOperatorAuthorization entity, vsoa.records in MOD-DE-MSG-5/6 and AUTHZ-CHECK-3 of the VPR spec. Renamed the public methods to spell out VsOperatorAuthorization anyway.

apps/vs-agent/src/admin.module.ts
tarunvadde: Removed

packages/agent-sdk/src/blockchain/AuthorizationService.ts
tarunvadde: Done

apps/vs-agent/src/main.ts
tarunvadde: Done

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
```diff
@@ -0,0 +1,62 @@
...
+// VSOA records change inside these pp msgs and, once https://github.com/verana-labs/verana-indexer/pull/324
+// lands, the forwarded de keeper events.
+const REFRESH_MSGS = [
+  'StartParticipantOP',
+  'SelfCreateParticipant',
+  'CreateRootParticipant',
+  'SetParticipantOPToValidated',
+  'SetParticipantEffectiveUntil',
+  'GrantVSOperatorAuthorization',
+  'UpdateVSOperatorAuthorization',
```
lotharking, line 15: The issue mentions `UpdateVSOperatorAuthorization`. Are you sure this is correct?

packages/agent-sdk/src/blockchain/AuthorizationService.ts
```diff
@@ -0,0 +1,103 @@
+import { BaseLogger } from '@credo-ts/core'
+
+import { VeranaChainService } from './VeranaChainService'
+import { CachedVsOperatorAuthorizationRecord, DurationParam } from './types'
+
+// A lapsed grant that carries a period is still valid: the chain rolls the expiration
+// forward on the next check (VPR spec AUTHZ-CHECK-3 step 4, AUTHZ-CHECK-1 step 2).
+function isActive(expiration?: Date, period?: DurationParam): boolean {
+  return expiration === undefined || expiration.getTime() > Date.now() || period != null
```
lotharking: Maybe it would be a good idea to validate `period != null` a bit more thoroughly.

packages/agent-sdk/tests/AuthorizationService.test.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  it('skips refreshes inside the configured interval', async () => {
+    const { chain, listVsOperatorAuthorizations } = makeChain()
+    const authz = new AuthorizationService({ chain, logger, minRefreshIntervalMs: 60_000 })
+    await authz.refreshForOperator()
+    await authz.refreshForOperator()
+    expect(listVsOperatorAuthorizations).toHaveBeenCalledTimes(1)
+  })
+
+  it('reports feegrant presence only for active with_feegrant records', async () => {
```
lotharking, line 136: > UpdateVSOperatorAuthorizationExpiration — refresh the cached record; recompute remaining feegrant validity when with_feegrant is true

I can see the `with_feegrant` validation. Is it still present?

packages/agent-sdk/src/blockchain/handlers/authorizationHandlers.ts
tarunvadde: Yes, verified against both ends. The chain's keeper event is `update_vs_operator_authorization` (`x/de/types/events.go`), and https://github.com/verana-labs/verana-indexer/pull/324 forwards it as `event_type` `UpdateVSOperatorAuthorization`, so that's the exact string that arrives on the wire. `UpdateVSOperatorAuthorizationExpiration` is the spec's method name (MOD-DE-MSG-9), not the emitted event name. The issue text used the spec name.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
tarunvadde: Done. Tightened it.

packages/agent-sdk/tests/AuthorizationService.test.ts
tarunvadde: Yes, still present. The expiration-update event triggers `refreshForOperator`, which rebuilds every cached record from current chain state, so `hasFeegrant` re-evaluates against the new expiration automatically. The test you're looking at pins that gating, and it now also covers the zero-period case.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
```diff
@@ -0,0 +1,104 @@
+import { BaseLogger } from '@credo-ts/core'
+
+import { VeranaChainService } from './VeranaChainService'
+import { CachedVsOperatorAuthorizationRecord, DurationParam } from './types'
+
+// A lapsed grant that carries a period is still valid: the chain rolls the expiration
+// forward on the next check (VPR spec AUTHZ-CHECK-3 step 4, AUTHZ-CHECK-1 step 2).
+function isActive(expiration?: Date, period?: DurationParam): boolean {
```
lotharking: I think we need an additional validation for `withFeegrant`. The current validation isn't sufficient because `hasFeegrant` requires different validation logic.

tarunvadde: hasFeegrant now checks expiration strictly, no period leniency. canSign keeps it since it mirrors AUTHZ-CHECK-3.

lotharking: Regarding `canSign`, it would probably be a good idea to also consider the expiration date, especially for participant registries.

tarunvadde: canSign did consider expiration, but through the operator-authorization semantics where no expiration means no expiry. For VSOA records the chain rejects a missing expiration (AUTHZ-CHECK-3 step 4), so the checks are now split: VSOA records require an expiration, operator authorizations keep the no-expiry case.

packages/agent-sdk/src/blockchain/AuthorizationService.ts
```diff
@@ -0,0 +1,119 @@
...
+    this.logger = config.logger
+    this.minRefreshIntervalMs = config.minRefreshIntervalMs ?? 2_000
+  }
+
+  async refreshForOperator(): Promise<void> {
+    // Catch-up replays hit the same current chain state, so back-to-back refreshes are skipped.
+    if (Date.now() - this.lastRefreshAt < this.minRefreshIntervalMs) return
+    this.lastRefreshAt = Date.now()
+
+    const vsoas = await this.chain.listVsOperatorAuthorizations()
```
lotharking: Probably better to update after the chain operation has been confirmed
```suggestion
    const vsoas = await this.chain.listVsOperatorAuthorizations()
    this.lastRefreshAt = Date.now()
```

lotharking (approved):

packages/agent-sdk/src/blockchain/AuthorizationService.ts
tarunvadde: Done

PR #496, opened 2026-07-09
feat(vs-agent): ECS credential bootstrap for standalone and delegated modes

tarunvadde:
Closes #476.

On startup the agent bootstraps its ECS credentials:
- **Standalone:** discovers the trusted ecosystem's ECS schemas via the indexer, then starts the HOLDER onboarding (org/persona) and the Service ISSUER onboarding on chain. Existing participants are reused, so restarts are idempotent. The issuer path follows the schema's `issuerOnboardingMode` (including `selfCreate` for OPEN).
- **Delegated:** verifies the parent VS, then requests the Service credential from it over vt-flow.

Offers are accepted through the vt-flow applicant pipeline (`autoAcceptCredentialOffer` + verify hook, now with retries for indexer lag).

New env: `TRUSTED_ECS_ECOSYSTEM_DIDS`, a comma-separated list of trusted ECS ecosystem DIDs per WL-ECS of the verifiable trust spec.

Validated with unit tests and an e2e against a live chain and indexer.

### DIFF
--- a/apps/vs-agent/src/config/constants.ts
+++ b/apps/vs-agent/src/config/constants.ts
@@ -144,7 +144,10 @@ export const VERANA_INDEXER_SUBSCRIPTION_SCOPE = (process.env.VERANA_INDEXER_SUB
   .toLowerCase()
 export const VERANA_AUTO_TRIGGER_RESOLVER = process.env.VERANA_AUTO_TRIGGER_RESOLVER !== 'false'
 
-export const TRUSTED_ECS_ECOSYSTEM_ID = process.env.TRUSTED_ECS_ECOSYSTEM_ID
+export const TRUSTED_ECS_ECOSYSTEM_DIDS = (process.env.TRUSTED_ECS_ECOSYSTEM_DIDS ?? '')
+  .split(',')
+  .map(did => did.trim())
+  .filter(Boolean)
 export const AGENT_MODE = (process.env.AGENT_MODE ?? '').trim().toLowerCase() || 'standalone'
 export const AGENT_DELEGATED_PARENT_VS_DID = process.env.AGENT_DELEGATED_PARENT_VS_DID
 
--- a/apps/vs-agent/src/config/constants.ts
+++ b/apps/vs-agent/src/config/constants.ts
@@ -144,6 +144,7 @@ export const VERANA_INDEXER_SUBSCRIPTION_SCOPE = (process.env.VERANA_INDEXER_SUB
   .toLowerCase()
 export const VERANA_AUTO_TRIGGER_RESOLVER = process.env.VERANA_AUTO_TRIGGER_RESOLVER !== 'false'
 
+export const TRUSTED_ECS_ECOSYSTEM_ID = process.env.TRUSTED_ECS_ECOSYSTEM_ID
 export const AGENT_MODE = (process.env.AGENT_MODE ?? '').trim().toLowerCase() || 'standalone'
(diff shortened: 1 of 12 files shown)


packages/agent-sdk/src/bootstrap/EcsBootstrapService.ts
```diff
@@ -0,0 +1,368 @@
...
+      connectionId,
+      schemaId: String(schemaId),
+      agentParticipantId: '0',
+      walletAgentParticipantId: '0',
+    })
+    this.logger.info(
+      `[EcsBootstrap] requested the Service credential (schema ${schemaId}) from parent VS ${parentDid}`,
+    )
+
+    await this.awaitDelegatedOutcome(record.id, parentDid)
```
lotharking, line 330: I have a misunderstanding here. Does `sendIssuanceRequest` wait for the operation to complete? If so, wouldn't the flow finish before this `await` is reached?

lotharking (changes requested):

packages/agent-sdk/src/bootstrap/EcsBootstrapService.ts
tarunvadde: No, it doesn't wait. `sendIssuanceRequest` persists the record in `IR_SENT`, dispatches the issuance-request message, and returns (it ends with `messageSender.sendMessage` and returns the record). Completion happens later, driven by the validator's inbound DIDComm messages. The listener in `awaitDelegatedOutcome` is registered synchronously right after the send returns, and inbound messages can't be processed in the middle of that, so there's no window where a state change could be missed.

apps/vs-agent/src/utils/setupAgent.ts
```diff
@@ -120,17 +120,25 @@ export const setupAgent = async ({
...
             }
-            try {
-              await orchestrator.verifyOfferedCredential(record.id)
-              return true
-            } catch (error) {
-              logger.error(`[vt-flow] credential verification failed: ${(error as Error).message}`)
-              return false
+            // Retries absorb indexer lag: the validator anchors the session and digest
+            // moments before offering, so the first reads can miss.
+            for (let attempt = 1; ; attempt++) {
```
lotharking: I'm not sure about this logic because, at some point, the `for` loop could end up blocking the service, and that could be risky.

apps/vs-agent/src/config/constants.ts
```diff
@@ -144,6 +144,10 @@ export const VERANA_INDEXER_SUBSCRIPTION_SCOPE = (process.env.VERANA_INDEXER_SUB
   .toLowerCase()
 export const VERANA_AUTO_TRIGGER_RESOLVER = process.env.VERANA_AUTO_TRIGGER_RESOLVER !== 'false'
 
+export const TRUSTED_ECS_ECOSYSTEM_DIDS = (process.env.TRUSTED_ECS_ECOSYSTEM_DIDS ?? '')
```
lotharking, line 147: Please update the docs

tarunvadde: Done

apps/vs-agent/src/utils/setupAgent.ts
tarunvadde: You might be right but It's bounded (10 attempts, 3s apart, around 30s on worst case) and it awaits async timers, so it only delays that one credential's verification, never the service. Restructured it to a visibly bounded loop so that's obvious from the code.

lotharking (approved):

PR #500, opened 2026-07-13
feat(vs-agent): corporation-scoped admin API auth (ADR-036)

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/472

Adds an authenticated external admin listener next to the existing internal one:

- `ADMIN_API_AUTH_MODE` picks the listeners: `internal` (unchanged, no auth) and/or `corporation` (new external listener on `ADMIN_API_EXTERNAL_PORT`).
- Auth flow: `POST /v1/auth/challenge` returns a nonce, the caller signs it with their Verana account key (ADR-036), `POST /v1/auth/token` verifies the signature offline and returns a short-lived bearer token.
- Routes declare an access mode: `PUBLIC`, `INTERNAL` (default, 403 on the external listener), or `CORPORATION` with the msg types the caller must hold. The guard checks the account's grant for the flow's validator participant through the authorization cache from #495.
- Swagger is disabled on the external listener.

### DIFF
--- a/apps/vs-agent/src/config/constants.ts
+++ b/apps/vs-agent/src/config/constants.ts
@@ -144,6 +144,10 @@ export const VERANA_INDEXER_SUBSCRIPTION_SCOPE = (process.env.VERANA_INDEXER_SUB
   .toLowerCase()
 export const VERANA_AUTO_TRIGGER_RESOLVER = process.env.VERANA_AUTO_TRIGGER_RESOLVER !== 'false'
 
+export const TRUSTED_ECS_ECOSYSTEM_DIDS = (process.env.TRUSTED_ECS_ECOSYSTEM_DIDS ?? '')
+  .split(',')
+  .map(did => did.trim())
+  .filter(Boolean)
 export const AGENT_MODE = (process.env.AGENT_MODE ?? '').trim().toLowerCase() || 'standalone'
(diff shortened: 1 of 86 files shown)


apps/vs-agent/src/config/constants.ts
```diff
@@ -29,6 +29,7 @@ export const PUBLIC_API_BASE_URL = process.env.PUBLIC_API_BASE_URL
 export const DEFAULT_PUBLIC_API_BASE_URL = 'http://localhost:3001'
 
 export const ADMIN_API_PUBLIC_URL = process.env.ADMIN_API_PUBLIC_URL
+export const ADMIN_API_EXTERNAL_PORT = Number(process.env.ADMIN_API_EXTERNAL_PORT || 3010)
```
lotharking, line 32: This hasn't been documented or included in the chart

apps/vs-agent/src/config/constants.ts
lotharking, line 1: > Annotate every existing admin route with its access mode

The issue mentions this, but the current PR only includes the new routes. What about the other admin routes?

lotharking (commented):
Only a few comments

apps/vs-agent/src/config/constants.ts
tarunvadde: Done

apps/vs-agent/src/config/constants.ts
tarunvadde: Done

genaris (approved):

PR #501, opened 2026-07-13
fix: validator-side issuance and agent startup bugs

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/499

- `op_summary_digest` is now omitted for HOLDER participants (MOD-PP-MSG-3) and the session is anchored with the validator participant as `issuer_participant_id`.
- `setupAgent` wires `autoIssueCredentialOnRequest`, so validators actually issue after the holder's request instead of stalling at `RequestReceived`.
- `VERANA_CORPORATION_ID` now resolves the corporation policy address from the indexer at startup and feeds it to the transaction signer.
- `indexerWs.start()` is guarded, so an agent with the indexer configured but no public DID and no corporation scope logs a warning instead of crashing.
- New `reconcileVtjscPublications` runs at startup and publishes any VTJSC missed while the agent was down or not yet subscribed.

All six were found and verified by running the full lifecycle against a live chain and indexer.

### DIFF
--- a/apps/vs-agent/src/main.ts
+++ b/apps/vs-agent/src/main.ts
@@ -10,8 +10,10 @@ import {
   VsAgentWsInboundTransport,
   type VsAgentNestPlugin,
   VeranaChainService,
+  VeranaIndexerService,
   IndexerWebSocketService,
   buildDefaultIndexerHandlerRegistry,
+  reconcileVtjscPublications,
 } from '@verana-labs/vs-agent-sdk'
 import * as express from 'express'
 import * as fs from 'fs'
@@ -240,13 +242,28 @@ const run = async () => {
     VtFlowNestPlugin,
   ]
 
+  const indexerService = VERANA_INDEXER_BASE_URL
+    ? new VeranaIndexerService({ baseUrl: VERANA_INDEXER_BASE_URL, logger: serverLogger })
+    : undefined
+
   // Connect to Verana blockchain for on-chain transactions
   let veranaChain: VeranaChainService | undefined
   if (VERANA_RPC_ENDPOINT_URL && VERANA_ACCOUNT_MNEMONIC) {
+    let corporationAddress: string | undefined
+    if (VERANA_CORPORATION_ID && indexerService) {
+      const corporation = await indexerService.getCorporation(VERANA_CORPORATION_ID).catch(() => undefined)
+      corporationAddress = corporation?.policy_address ?? undefined
+      if (!corporationAddress) {
+        serverLogger.warn(
+          `Corporation ${VERANA_CORPORATION_ID} not resolvable on the indexer yet; on-chain transactions will sign without a corporation`,
+        )
+      }
+    }
     veranaChain = new VeranaChainService({
       rpcUrl: VERANA_RPC_ENDPOINT_URL,
       chainId: VERANA_CHAIN_ID,
       mnemonic: VERANA_ACCOUNT_MNEMONIC,
+      corporationAddress,
       logger: serverLogger,
       autoTriggerResolver: VERANA_AUTO_TRIGGER_RESOLVER,
     })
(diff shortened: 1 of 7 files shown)


packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -322,6 +323,37 @@ export async function startParticipantOPAutoFlow(agent: VsAgent, activity: Index
...
+  if (!agent.did || !agent.publicApiBaseUrl) return
+
+  const [didRecord] = await agent.dids.getCreatedDids({ did: agent.did })
+  if (!didRecord) return
+  const chainId = agent.veranaChain?.getChainId ?? DEFAULT_CHAIN_ID
+
+  const ecosystems = await indexer.listEcosystems()
+  for (const ecosystem of ecosystems.filter(entry => Number(entry.corporation_id) === corporationId)) {
+    for (const schema of await indexer.listCredentialSchemas(ecosystem.id)) {
+      if (getTrustMetadata(didRecord, '_vt/jsc', String(schema.id))) continue
```
lotharking: Please review the `getTrustMetadata` function. In my opinion, this looks like a workaround rather than the right solution.

There's another issue as well. I don't think this implementation works correctly. Normally, `schema.id` is used as the key to map the value, but in this implementation the key is `jsonSchemaRef`. Because of that, I think you should use:

`await findMetadataEntry(didRecord, '_vt/jsc', '', schemaRef)`

apps/vs-agent/src/main.ts
```diff
@@ -240,13 +242,28 @@ const run = async () => {
...
 
+  const indexerService = VERANA_INDEXER_BASE_URL
+    ? new VeranaIndexerService({ baseUrl: VERANA_INDEXER_BASE_URL, logger: serverLogger })
+    : undefined
+
   // Connect to Verana blockchain for on-chain transactions
   let veranaChain: VeranaChainService | undefined
   if (VERANA_RPC_ENDPOINT_URL && VERANA_ACCOUNT_MNEMONIC) {
+    let corporationAddress: string | undefined
+    if (VERANA_CORPORATION_ID && indexerService) {
```
lotharking, line 253: Please keep in mind that if the corporation ID is present and this setup is expected to fail without it, we probably don't need to support a WebSocket subscription configuration without a corporation ID.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
tarunvadde: Thanks for pointing it out. Using `schemaRef ` now.

apps/vs-agent/src/main.ts
tarunvadde: Corporate agents do need the corporation id now, agreed. But the DID-scoped subscription without one has to stay: applicant-side and delegated agents run without any corporation (a delegated child is even chainless), and they still consume events about their own participants, like RevokeParticipant for credential cleanup. The corporation id is required for agents that act on behalf of one, not for agents that only hold credentials.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -322,6 +323,38 @@ export async function startParticipantOPAutoFlow(agent: VsAgent, activity: Index
...
+  corporationId: number,
+): Promise<void> {
+  if (!agent.did || !agent.publicApiBaseUrl) return
+
+  const [didRecord] = await agent.dids.getCreatedDids({ did: agent.did })
+  if (!didRecord) return
+  const chainId = agent.veranaChain?.getChainId ?? DEFAULT_CHAIN_ID
+
+  const ecosystems = await indexer.listEcosystems()
+  for (const ecosystem of ecosystems.filter(entry => Number(entry.corporation_id) === corporationId)) {
```
lotharking, line 338: I'm not sure this is entirely correct because a corporation ID can be associated with multiple ecosystems.

tarunvadde: It is a filter, not a find, so every matching ecosystem and its schemas are walked. Verified live: added a second ecosystem while the agent was down, on boot it published exactly the one missing VTJSC and skipped the rest.

If you mean the agent should not publish for all of its corporation's ecosystems: `publishVtjscIfOwner` has no ownership check either, so a stricter rule belongs to both paths and its own issue.

lotharking (approved):

PR #506, opened 2026-07-14
fix(vt-flow): credential cleanup on revocation and cancel-aware flow reconciliation

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/503

- New `onCredentialRevoked` hook: when an applicant flow enters `CRED_REVOKED`, the agent removes the credential's LinkedVerifiablePresentation entry from its DID Document and deletes the credential from the wallet, as the spec requires.
- `CancelParticipantOPLastRequest` no longer terminates every flow: if the participant is still VALIDATED on chain (a renewal was cancelled), flows reset by that renewal restore to `COMPLETED`. Cancelling a first request still terminates.

### DIFF
--- a/packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
+++ b/packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
@@ -6,15 +6,16 @@ import {
   VtFlowRole,
   VtFlowService,
   VtFlowState,
+  isVtFlowTerminalState,
 } from '@verana-labs/credo-ts-didcomm-vt-flow'
 import { computeSchemaDigest } from '@verana-labs/vs-agent-model'
 
 import { VsAgent } from '../../agent/VsAgent'
 import { getEcsSchemas } from '../../utils/data'
-import { createJsc, findMetadataEntry, removeTrustCredential } from '../../utils/trustCredentialStore'
+import { createJsc, findMetadataEntry, removeStoredTrustCredential } from '../../utils/trustCredentialStore'
 import { VtFlowOrchestrator } from '../../vtFlow'
 import { VeranaIndexerService } from '../VeranaIndexerService'
-import { IndexerActivity, VeranaSyncState } from '../types'
+import { IndexerActivity, ValidationState, VeranaSyncState } from '../types'
 
 const DEFAULT_CHAIN_ID = 'vna-testnet-1'
 const PARTICIPANT_ROLE_HOLDER = 6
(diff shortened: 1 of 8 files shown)


packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -291,18 +291,25 @@ export async function removeHolderTrustCredentialIfRevoked(
...
 
-export async function terminateVtFlowRecordsByApplicant(
-  agent: VsAgent,
-  participantId: string,
-): Promise<void> {
+const ONBOARDING_STATE_VALIDATED = 2
+
+export async function reconcileVtFlowRecordsOnCancel(agent: VsAgent, participantId: string): Promise<void> {
+  const participant = await agent.veranaChain?.getParticipant(Number(participantId)).catch(() => undefined)
+  const stillValidated = Number(participant?.opState) === ONBOARDING_STATE_VALIDATED
```
```suggestion
  const stillValidated = Number(participant?.opState) === ValidationState.VALIDATED
```

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -255,6 +255,27 @@ export class VtFlowOrchestrator {
     }
   }
 
+  async onCredentialRevoked(vtFlowRecordId: string): Promise<void> {
```
lotharking, line 259: This method is very similar to `removeHolderTrustCredentialIfRevoked`. Please review it.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -291,18 +291,25 @@ export async function removeHolderTrustCredentialIfRevoked(
...
+
   await reconcileVtFlowRecordsForParticipant(
     agent,
     participantId,
     async (record, service, agentContext) => {
+      const renewalState = record.state === VtFlowState.AwaitingOr || record.state === VtFlowState.OrSent
+      if (stillValidated && renewalState && record.credentialExchangeRecordId) {
+        await service.updateState(agentContext, record, VtFlowState.Completed)
+        return 'COMPLETED'
+      }
```
lotharking, line 301: Should this branch on op_state alone? Per VSA-VTI-FLOW-OP-CANCEL, VALIDATED → COMPLETED. With the renewalState guard, cancelling a renewal at OOB_PENDING/VALIDATING terminates it instead.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
tarunvadde: Done

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
tarunvadde: Consolidated both onto a shared `removeStoredTrustCredential` helper. Note this also fixes the chain-event path: it only removed the linked VP before, and MOD-PP-MSG-9 also requires deleting the credential from the store.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
tarunvadde: You are right ig. It now branches on op_state plus a prior credential exchange, so OOB_PENDING and VALIDATING restore to COMPLETED too, and terminal or already-completed records are left untouched.

lotharking (approved):
I like it. I only have one suggestion: since we don't have a way to test this manually at the moment, could you please add some tests? Unit tests would be fine. The goal is to verify that the mapping behaves as expected and covers the main scenarios.

PR #507, opened 2026-07-14
fix: v4 conformance details for VTJSC naming, managed mutations, and applicant roles

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/504

- VTJSC DID Document service fragments now end with `-vtjsc-vp` per VT-ECOSYSTEM-DIDDOC.
- Managed mutations respond `409 Conflict` instead of 403, and the guard now also covers `POST/DELETE /v1/vt/linked-credentials`.
- `startOnboardingProcess` accepts VERIFIER, ISSUER_GRANTOR, and VERIFIER_GRANTOR applicants.
- `agent_participant_id` / `wallet_agent_participant_id` are configurable through orchestrator options (default 0).


lotharking (commented):
I have a couple of questions.

The first one is about `wallet_agent_participant_id` and `agent_participant_id`. This PR doesn't seem to include any changes related to them. What is the plan for those fields?

The second one is about the legacy persistence. Remember that the current flow checks the digest of the self credentials and recreates them if needed. At the moment, this implementation only updates the existing records, so I'm not sure how that scenario is handled.

I don't think removing the legacy implementation is necessarily a problem, but I'd like to understand the proposed solution. Could you explain the approach and add a comment about it?

lotharking (approved):

PR #508, opened 2026-07-14
fix(v1.11): repair webvh update-key mapping so public DID self-heal can publish

genaris:
## Problem

In some cases (like X-Agent Hologram Verifiable Service demo), after upgrading to v1.11.0, a peer connecting to a `did:webvh` vs-agent via **implicit DIDComm v1 DID Exchange** is rejected with:

> DID Rotate signature is missing.

The agent's `did-communication` service advertises `recipientKeys` pointing at the X25519 key-agreement key instead of the Ed25519 key, so Credo builds the DID Exchange response's `did_rotate~attach` with an empty signing-key set — the attachment has a payload but no JWS.

The `recipientKeys` generation is already correct in current code; the agent is *supposed* to fix an old document on startup via its self-heal (`servicesChanged`) path. But on affected agents that self-heal **silently fails to persist**.

## Root cause

For `did:webvh`, updating the DID document means signing a new log entry with the DID's **update key**. `WebVhDidRegistrar.update()` resolves the update key's `kmsKeyId` via `getKmsKeyIdForVerifiacationMethod`, which matches by suffix:

```js
keys.find(({ didDocumentRelativeKeyId }) => vm.id.endsWith(didDocumentRelativeKeyId))
```

Records created by older versions stored the update key's `didDocumentRelativeKeyId` as the full multibase (`#z6Mk…BVhGnL79`), while the DID document VM uses a short fragment (`#BVhGnL79`). `"…#BVhGnL79".endsWith("#z6Mk…BVhGnL79"`) is false, so the update key is unresolvable and the registrar returns `state:"failed"` with:

```
The key ID must be present before the log can be edited.
```

The update result was never checked, so this was logged as success while the document never changed — which is why restarting never helped. The private key is still present in the KMS; only the mapping label is wrong.

## Fix
`repairWebvhUpdateKeyMapping()` runs before the self-heal update; rewrites a webvh key mapping whose relative id no longer correlates to any VM back to the VM's actual fragment. Idempotent, and a no-op for healthy records.
Check the `dids.update()` result: log the failure reason instead of assuming success, so a failed webvh update can't masquerade as a successful one again.

Healthy agents are unaffected (repair is a no-op; result check only adds an error log on genuine failures).

Note: this is for **v1.11** (to release a v1.11.1 hotfix). Another PR targetting main will be created afterwards.


packages/agent-sdk/src/agent/VsAgent.ts
```diff
@@ -353,6 +365,40 @@ export class VsAgent<TModules extends BaseAgentModules = BaseAgentModules> exten
...
+          to: correct,
+          kmsKeyId: key.kmsKeyId,
+        })
+        key.didDocumentRelativeKeyId = correct
+        repaired = true
+      }
+    }
+    if (repaired) {
+      const didRepository = this.dependencyManager.resolve(DidRepository)
+      await didRepository.update(this.agentContext, record)
```
lotharking, line 398: I recommend check the response in this update method.

packages/agent-sdk/src/agent/VsAgent.ts
```diff
@@ -353,6 +365,40 @@ export class VsAgent<TModules extends BaseAgentModules = BaseAgentModules> exten
...
+      const vm = vms.find(v => `#${v.publicKeyMultibase}` === rel) // was stored as #<multibase>
+      if (!vm) continue
+      const correct = `#${vm.id.split('#')[1]}`
+      if (correct !== rel) {
+        this.logger?.warn('Fixing webvh update-key mapping', {
+          from: rel,
+          to: correct,
+          kmsKeyId: key.kmsKeyId,
+        })
+        key.didDocumentRelativeKeyId = correct
```
lotharking, line 392: The new Credo version uses metadata to persist this key. It would probably be a good idea to update it.

packages/agent-sdk/src/agent/VsAgent.ts
genaris: didRepository.update returns Promise<void>. It will throw in case there is an error while updating DidRecord. Since this will be shown in vs-agent log, I don't see so much benefits on adding a specific try-catch to it (also it is not likely to throw, since the record has been fetched some lines above).

packages/agent-sdk/src/agent/VsAgent.ts
genaris: Well, I didn't want to make this fix more complex but probably it is fine to do a version bump here.

lotharking (approved):

tarunvadde (approved):

PR #510, opened 2026-07-15
feat(agent-sdk): direct issuance validator side

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/502

- New `buildDirectIssuanceOffer` on the orchestrator: resolves the agent's own ACTIVE ISSUER participant for the requested schema via the indexer, anchors `CreateOrUpdateParticipantSession` with the credential digest, and returns the signed credential offer. Returns nothing when the agent holds no matching issuer participant, so unknown schemas are not auto-offered.
- `setupAgent` wires `autoAcceptIssuanceRequest`, `autoOfferCredential`, and `buildCredentialOffer`, so a production validator now answers inbound `issuance-request` messages.
- The auto-offer hook passes `credentialDigest` through to `offerCredentialForSession`, so the offered digest is persisted on the flow record.
- The issuer participant is selected by matching vs_operator to the agent's operator account, since the chain rejects session anchoring signed by anyone else.

This makes delegated mode (#496) work end to end: validated on a live chain and indexer, with a delegated child bootstrapping against a vs-agent parent and receiving its Service credential over DIDComm, session digest verified on the indexer.


packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -211,6 +200,96 @@ export class VtFlowOrchestrator {
...
+    subjectDid: string
+    claims: JsonObject
+    credentialType?: string[]
+    credentialContext?: string[]
+  }): Promise<{ unsignedCredentialJson: JsonCredential; digest: string }> {
+    const chain = this.requireChain()
+    const didRecords = await this.agent.dids.getCreatedDids({ did: this.agent.did! })
+    const didRecord = didRecords[0]
+    if (!didRecord) throw new Error('Agent DID record not found')
+    const schemaRef = `vpr:verana:${chain.getChainId}/cs/v1/js/${input.credentialSchemaId}`
```
lotharking: It would probably be a good idea to use the new path here, right?

lotharking (commented):
Only a comment

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
tarunvadde: Done.

lotharking (approved):

PR #512, opened 2026-07-15
fix: deploy only two vs-agent flavours

genaris:
Creating now only 2 different vs-agent container flavors, to maximize backwards compatibility (99% of existing services) and not make things more complicate than needed for now:

- veranalabs/vs-agent: keep all functionality except eMRTD
- veranalabs/vs-agent-mrtd: add eMRTD (used in passport demos)


tarunvadde (approved):
LGTM.

PR #514, opened 2026-07-15
feat(chart): configurable redis image, maxmemory and extra args

mjfelis:
## Overview

The vs-agent-chart's Redis container was fully hardcoded: floating `redis:alpine` image, no way to pass any Redis configuration. In particular `maxmemory` could not be set, so Redis defaults to unlimited memory and only discovers the pod's cgroup limit when the kernel OOM-kills it.

New values:

| Value | Description | Default |
|---|---|---|
| `redis.image` | Redis container image (pinnable) | `redis:alpine` |
| `redis.maxmemory` | Redis `maxmemory`; set `""` for unlimited | **`80mb`** |
| `redis.maxmemoryPolicy` | `maxmemory-policy`, applied only when `maxmemory` is set | `noeviction` |
| `redis.extraArgs` | Additional `redis-server` flags (list) | `[]` |

**Behavior change**: the default is now a **capped Redis (80mb, noeviction)** instead of unlimited — sized for the queue-only role this Redis plays in vs-agent (Bull `message` queue), and a fit for the current deployments (avatar/passport run well under 80 MB). Deployments needing more can raise `redis.maxmemory` or set `""` to restore the old unlimited behavior. `noeviction` is deliberate: a job queue must fail loudly rather than have keys silently evicted.

Rendered args go through the official redis image entrypoint (flag-style args are prepended with `redis-server` automatically):

```yaml
args: ["--maxmemory", "80mb", "--maxmemory-policy", "noeviction"]
```

Verified with helm 3.16: `helm lint` clean; default values render the 80mb cap; `maxmemory: ""` renders no args (identical to previous behavior); `redis.enabled=false` unaffected.

Related: #515 fixes the Redis PVC mount path so persistence actually works.

Context: the AI agents' Redis OOM crash-loops in hologram-verifiable-services (see 2060-io/hologram-verifiable-services#63) — this chart change is the prerequisite for capping the standalone vs-agent Redis instances (avatar/passport) at a small `maxmemory`.

### DIFF
--- a/charts/vs-agent/README.md
+++ b/charts/vs-agent/README.md
@@ -92,7 +92,7 @@ extraEnv:
 | `redis.host`               | Redis host                                       | `your-redis-host`               |
 | `redis.password`           | Redis password                                   | `myRedisPass123`                |
 | `redis.image`              | Redis container image (pin a tag for reproducible deploys) | `redis:alpine`       |
-| `redis.maxmemory`          | Redis `maxmemory` (e.g. `80mb`); empty = unlimited | `""`                          |
+| `redis.maxmemory`          | Redis `maxmemory`; set `""` for unlimited        | `80mb`                          |
 | `redis.maxmemoryPolicy`    | Redis `maxmemory-policy` (applied only when `maxmemory` is set) | `noeviction`     |
 | `redis.extraArgs`          | Additional `redis-server` flags (list)          | `[]`                            |
 
--- a/charts/vs-agent/README.md
+++ b/charts/vs-agent/README.md
@@ -91,6 +91,10 @@ extraEnv:
 | `redis.enabled`            | Enable Redis                                     | `false`                         |
 | `redis.host`               | Redis host                                       | `your-redis-host`               |
 | `redis.password`           | Redis password                                   | `myRedisPass123`                |
+| `redis.image`              | Redis container image (pin a tag for reproducible deploys) | `redis:alpine`       |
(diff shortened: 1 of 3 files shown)


genaris (commented):
@lotharking it would be nice if, besides these parameters, is anything we should also tweak based on the usage of redis we really do on VS Agent, since I see we use defaults for everything and maybe we are using more resources than actually needed.

charts/vs-agent/README.md
```diff
@@ -91,6 +91,10 @@ extraEnv:
 | `redis.enabled`            | Enable Redis                                     | `false`                         |
```
genaris, line 91: Something that I don't get in this chart configuration is how redis.host and redis.password work, since "redis.enabled" is meant to enable or disable the **internal** redis container (i.e. don't use an external one). 

This PR is adding several redis.* parameters, all of them related to the internal redis container.

@lotharking is my reasoning correct? How could we make it clear that all of this is for the internal one, and how at the same time allow to use an external redis (and properly document these deployment use cases)?

lotharking: These chart values aren't wired up yet: redis.host and redis.password appear in values.yaml and the docs, but the template doesn't use them, so the internal Redis config doesn't rely on them at the moment.

genaris: Then I think it is better to remove them... they will only cause confusion. But it can be left for a further PR.

genaris (approved):

PR #521, opened 2026-07-21
chore: migrate from eslint and prettier to biome

tarunvadde:
Closes [#519](https://github.com/verana-labs/vs-agent/issues/519).


examples/nestjs-vs/package.json
```diff
@@ -8,14 +8,11 @@
   "scripts": {
     "build": "nest build && pnpm run copy-i18n",
     "copy-i18n": "cp -r ./src/i18n ./build/i18n",
-    "check-format": "prettier -c \"src/**/*.ts\"",
-    "check-types": "eslint \"{src,apps,libs,test}/**/*.ts\"",
-    "format": "prettier --write \"src/**/*.ts\"",
+    "check-types": "tsc --noEmit",
```
lotharking, line 11: Why remove these scripts? This project is intended to make it easy to copy, paste, and quickly create chatbots. We should keep a copy of this format here rather than just removing the legacy version.

genaris: I think this is because the formatter runs directly at the root of the workspace.

package.json
```diff
@@ -18,26 +18,19 @@
     "start:verifier": "pnpm --filter demo-verifier start",
     "start:nestjsvs": "pnpm --filter nestjs-vs start",
     "check-types": "pnpm -r --workspace-concurrency=1 exec tsc --noEmit",
-    "format": "prettier \"{packages,apps}/**/*.ts\" --write",
-    "check-format": "prettier -c \"{packages,apps}/**/*.ts\"",
+    "format": "biome format --write .",
```
genaris:
```suggestion
    "fix-format": "biome format --write .",
```

This is to match the ts-project-template. It will need to change also in CI/CD scripts I think.

package.json
```diff
@@ -18,26 +18,19 @@
...
+    "check-format": "biome check .",
     "test": "pnpm -r run test",
     "test:e2e": "pnpm --filter @verana-labs/vs-agent-sdk test:e2e",
-    "lint": "eslint \"{packages,apps}/**/*.ts\" --fix",
+    "lint": "biome lint --write .",
```
genaris, line 25: I guess we don't have an exact equivalent in the template for this, but it is interesting. Would this be a sort of 'fix-types'? Maybe it could be good to add it on https://github.com/2060-io/ts-project-template/pull/3/. @lotharking WDYT?

genaris (approved):
LGTM, but I'd suggest to merge after it matches with the ts-project-template once https://github.com/2060-io/ts-project-template/pull/3/ is merged.

examples/nestjs-vs/package.json
tarunvadde: Yes. Ariel is right and on the copy-paste goal, a local biome config would not actually help. The folder cannot be copied out today regardless, because `@verana-labs/vs-agent-nestjs-client` and `@verana-labs/vs-agent-plugin-mrtd` are `workspace:*` deps, so `pnpm install fails` immediately in a copied folder. That has been true since before this PR.

genaris (approved):

PR #523, opened 2026-07-27
fix(agent-sdk): spec-compliant ECS publication and VTJSC digestSRI

tarunvadde:
Agents published linked VPs under legacy -c-vp fragments, hashed VTJSC digestSRI over a canonicalized form, and left self-issued ECS VPs bound to the self-TR example JSC. All three break trust resolution against a spec-following resolver.

- publish linked-VP fragments as vpr-schemas-*-vtc-vp per [VT-CRED-W3C-LINKED-VP]
- compute digestSRI over the raw on-chain schema bytes (SRI semantics)
- rebind self-issued ECS VPs to the on-chain VTJSC when the schema digest-identifies as an official ECS, and link the VP as a DID document service

Validated end to end against a live chain, indexer and verre (trusted resolution with real ECS credentials). Needs a verre release with verana-labs/verre#118 consumed by the indexer before deployment: agents publishing the spec fragments are only resolvable once the resolver matches them.


packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -25,6 +29,11 @@ import { IndexerActivity, ValidationState, VeranaSyncState } from '../types'
 const DEFAULT_CHAIN_ID = 'vna-testnet-1'
 const PARTICIPANT_ROLE_HOLDER = 6
 
+// digestSRI over the schema exactly as the registry serves it; consumers (verre) verify raw bytes
+function rawSchemaDigestSri(rawJsonSchema: string): string {
+  return `sha384-${createHash('sha384').update(rawJsonSchema).digest('base64')}`
```
lotharking: This is duplicated logic; please review it.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
```diff
@@ -25,6 +29,11 @@ import { IndexerActivity, ValidationState, VeranaSyncState } from '../types'
 const DEFAULT_CHAIN_ID = 'vna-testnet-1'
 const PARTICIPANT_ROLE_HOLDER = 6
 
+// digestSRI over the schema exactly as the registry serves it; consumers (verre) verify raw bytes
+function rawSchemaDigestSri(rawJsonSchema: string): string {
```
```suggestion
function rawSchemaDigestSRI(rawJsonSchema: string): string {
```

packages/agent-sdk/src/utils/trustCredentialStore.ts
```diff
@@ -344,3 +348,64 @@ export async function migrateVtjscServiceIds(agent: VsAgent): Promise<void> {
...
+// replaces the self-TR example JSC binding with the on-chain VTJSC so resolvers can link the credential to the VPR
+export async function rebindEcsCredentialSchema(
+  agent: VsAgent,
+  publicApiBaseUrl: string,
+  schemaId: string,
+  schemaKey: string,
+  defaults: SelfTrDefaults,
+): Promise<void> {
+  if (!['ecs-service', 'ecs-org'].includes(schemaKey) || !agent.did) return
+  const vpUrl = `${publicApiBaseUrl}/vt/${schemaKey}-c-vp.json`
```
lotharking: I'm not sure, but I think this is the correct format. Please review it. Also, I'm not sure if c-vp is the correct way to build the URL.
```suggestion
  const vpUrl = `${publicApiBaseUrl}/vt/schemas-${schemaKey}-c-vp.json`
```

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
tarunvadde: Thanks for pointing it out.

packages/agent-sdk/src/blockchain/handlers/stateMutations.ts
tarunvadde: Folded into the comment above. Rather than rename it I dropped the helper and called generateDigestSRI, which is the same computation.

packages/agent-sdk/src/utils/trustCredentialStore.ts
tarunvadde: I looked into this and I think the current URL is fine. The spec only says the DID service fragment has to start with `#vpr-schemas-` and end with `-vtc-vp`. There is no rule about the file name. The `schemas-example-vtc-vp.json` you have probably seen sits in a non-normative example, and verre only matches the fragment and fetches the endpoint as it is.

The URL also has to stay the same as the three other places that build it, otherwise the rebind stops covering those services.

You were right that something was off nearby though. VsAgent was emitting `#vpr-ecs-service-c-vp` while setupSelfTr and rebindEcsCredentialSchema used `#vpr-schemas-service-vtc-vp`. Since rebind appends its own service without removing the other one, DID documents ended up with both pointing at the same VP and verre resolved it twice. Fixed and both use the spec form now.

lotharking (dismissed):
Hey @genaris , nothing from my side.

packages/agent-sdk/package.json
```diff
@@ -47,7 +47,7 @@
     "@openwallet-foundation/askar-nodejs": "^0.6.0",
     "@verana-labs/credo-ts-didcomm-vt-flow": "workspace:*",
     "@verana-labs/verana-types": "0.10.1-dev.16",
-    "@verana-labs/verre": "0.3.2-dev.1",
+    "@verana-labs/verre": "0.3.3-dev.2",
```
genaris, line 50: Can't we just bump to 0.4.0-dev.X?

genaris: Oh, I see now that dev are still under 0.3.3. We should fix the CI on verre.

tarunvadde: Yess

genaris (approved):

PR #526, opened 2026-07-28
fix: migrate ECS credentials, schema routing and linked VP naming to v4

tarunvadde:
Finishes the v4 migration for ECS credentials and drops the remaining v3 paths.

- v4 [ECS-EC] digests plus the missing ecs-badge entry
- schema refs resolve to `/v4/credential-schema/js/{id}`, the old `/verana/cs/v1/js/ `output and input forms are gone
- linked VP files renamed to match their fragments, `-c-vp.json` to `-vtc-vp.json` and `-jsc-vp.json` to `-vtjsc-vp.json`
- verre bumped to 0.3.3-dev.2
- Two things worth calling out. The `/vt/:schemaId` route dispatches on the filename suffix, so it changed with the rename and agents need redeploying rather than self-healing. And I removed the legacy `_vt/jsc` ref handling on the assumption there is no v3 to v4 in place upgrade, happy to put it back if that is wrong.


packages/model/src/utils/ecs.ts
genaris: Note that this URL map does not work properly, since V4 indexer does not include the `/verana/` path

packages/model/src/utils/ecs.ts
```diff
@@ -22,11 +22,12 @@ export function mapToEcosystem(input: string): string {
   return input
```
genaris: Unless it is useful for particular things like self-issued credentials, we can drop the former form (/cs/v1/js/xxx)

packages/model/src/utils/ecs.ts
```diff
@@ -22,11 +22,12 @@ export function mapToEcosystem(input: string): string {
...
+// v4 spec [ECS-EC] reference digests
 export const ECS_SCHEMA_DIGESTS: Record<ECS, string> = {
-  [ECS.SERVICE]: 'sha384-PVseqJJjEGMVRcht77rE2yLqRnCiLBRLOklSuAshSEXK3eyITmUpDBhpQryJ/XIx',
-  [ECS.ORG]: 'sha384-XF10SsOaav+i+hBaXP29coZWZeaCZocFvfP9ZeHh9B7++q7YGA2QLTbFZqtYs/zA',
-  [ECS.PERSONA]: 'sha384-4vkQl6Ro6fudr+g5LL2NQJWVxaSTaYkyf0yVPVUmzA2leNNn0sJIsM07NlOAG/2I',
-  [ECS.USER_AGENT]: 'sha384-yLRK2mCokVjRlGX0nVzdEYQ1o6YWpQqgdg6+HlSxCePP+D7wvs0+70TJACLZfbF/',
+  [ECS.SERVICE]: 'sha384-0v+BAFGpnBX/RVqH9dUlMglxMrD4AKy4qUtb1lMN4iW9I2gO7XjcUfmGOf0oInP3',
+  [ECS.ORG]: 'sha384-UPn4TDqS1nMBAN3FyMzTAZOWp99zBjBD69OjpbhwOKZj7iOrS5qPwJ2SArRz0yzu',
+  [ECS.PERSONA]: 'sha384-VfXTfuks02OkoR5USaTfEdc4NU25m4+vNrLATnjC0r0Pn1S3tFTdOvGCfSYdjE2I',
+  [ECS.USER_AGENT]: 'sha384-rIWkh3zBD1Ak7CNGpAwZ/ONSmf+ywOYSF3H60ULc9/a1ZYKv6EqiQMJ2dm8dOfjm',
```
genaris, line 25: There is also an ECS.BADGE. See https://verana-labs.github.io/verifiable-trust-spec/versions/v4/#ecs-ec-essential-credential-schemas-ecosystem

packages/model/src/utils/ecs.ts
tarunvadde: Thanks for pointing it out. I also removed the `vna-mainnet-1` entry. It pointed at the testnet host and `idx.mainnet.verana.network` does not resolve, so mainnet refs were getting answered with testnet data. They fail now instead of quietly returning the wrong thing.

packages/model/src/utils/ecs.ts
tarunvadde: Added. While I was in there I bumped verre to 0.3.3-dev.2. We were on 0.3.2-dev.1, which still carries the v3 digests and no badge, so the agent and the resolver we ship were disagreeing about what counts as an ECS.

packages/model/src/utils/ecs.ts
tarunvadde: Half of it is gone. The function no longer rewrites a canonical `:cs:` ref into that shape first, so v4 networks never touch it.

I kept it for v3 networks. testnet only serves `/verana/cs/v1/js/{id}` today and 404s on the v4 route, so dropping it breaks testnet right now. It is also still MOD-CS-QRY-3 in the VPR v4 spec. The entry goes away on its own once testnet upgrades.

Self-issued credentials use absolute URLs, so they pass through untouched either way.

Separate thing worth raising though the VPR v4 spec never mentions /v4/credential-schema anywhere, so the spec and the indexer disagree about this path. Might be worth an issue, what do you think?

packages/agent-sdk/src/utils/setupSelfTr.ts
```diff
@@ -107,35 +108,44 @@ export const setupSelfTr = async ({
...
-      {
-        id: mapToSelfTr(schemaUrl, publicApiBaseUrl),
-        type: 'JsonSchemaCredential',
-      },
-      defaults,
-    )
+    try {
+      await generateVerifiablePresentation(
+        agent,
+        `${publicApiBaseUrl}/vt/${name}-c-vp.json`,
```
lotharking:
```suggestion
        `${publicApiBaseUrl}/vt/${name}-vtc-vp.json`,
```

packages/agent-sdk/src/utils/setupSelfTr.ts
```diff
@@ -515,44 +527,12 @@ export function generateDigestSRI(content: string, algorithm: string = 'sha384')
...
-        `Failed to convert URL to Base64. URL: ${url}. ` +
-          `Status: ${error.response?.status ?? 'N/A'}. ` +
-          `Message: ${error.message}`,
-      )
-    } else {
-      logger.error(`Unexpected error converting URL to Base64: ${error}`)
-    }
-    return fallback
-  }
+export async function urlDigestSri(url: string): Promise<string> {
```
lotharking, line 533: Hey @tarunvadde, I have a question about this. In Verre and some other places, we calculate the digest by removing the schema ID rather than directly from the URL. This is a legacy approach that was recommended before.
I'm not sure if the spec mentions anything about this, because changing it would require updates to other services as well.

lotharking (approved):

lotharking (changes requested):
I only approved it by mistake. I meant to leave it unapproved for now.

packages/model/src/utils/ecs.ts
genaris: > I kept it for v3 networks. testnet only serves /verana/cs/v1/js/{id} today and 404s on the v4 route, so dropping it breaks testnet right now. It is also still MOD-CS-QRY-3 in the VPR v4 spec. The entry goes away on its own once testnet upgrades.

vs-agent v2.0.0 is not meant to be compatible with current testnet (v3), so that is not a problem. `MOD-CS-QRY-3` is part of verana-node API (implementation of VPR), but not the indexer, which follows a completely different API (that's why `/v4/*` are not mentioned in VPR specification). In my opinion, this is confusing and, at some point, we should drop the VPR query API methods and define the ones from the indexer directly. For the time being, blockchain query methods are useful to do some checks and verify indexer behaviour while developing.

The important concept here is that neither vs-agent, verana-graph, hologram or any other app should be calling the blockchain query API directly, but the one of the indexer.

packages/agent-sdk/src/utils/setupSelfTr.ts
genaris: Digest SRI is not the same as what we call "DigestJCS". It follows a another specification, widely used in other use cases: https://www.w3.org/TR/sri-2/

packages/model/src/utils/ecs.ts
tarunvadde: You are right, dropped it entirely. Only the canonical `:cs:` form now, always the v4 route.

Also retracting my point about the spec and the indexer disagreeing, that was me misreading MOD-CS-QRY-3. I removed the legacy `_vt/jsc` handling too, assuming no v3 to v4 in place upgrade. Say if that is wrong.

packages/agent-sdk/src/utils/setupSelfTr.ts
tarunvadde: Done, and did the schema VPs too so both families match their fragments.

It is consistency rather than compliance though, the spec only makes the fragment normative. Your suggestion did surface a real one: the `/vt/:schemaId `route dispatches on the filename suffix, so renaming alone would have 400d every VP fetch. Fixed.

apps/vs-agent/src/controllers/admin/verifiable/TrustController.ts
```diff
@@ -133,7 +135,7 @@ export class TrustController {
       verifiableTrustCredential: {
         summary: 'Verifiable Trust Credential example',
         description: 'A full URL to the Verifiable Trust Credential.',
-        value: 'https://p2801.ovpndev.mobiera.io/vt/ecs-service-c-vp.json',
+        value: 'https://p2801.ovpndev.mobiera.io/vt/ecs-service-vtc-vp.json',
```
genaris: All these p2801.ovpndev.mobiera.io are not good examples. Change by myservice.verana.io or similar.

apps/vs-agent/src/config/constants.ts
```diff
@@ -110,16 +110,14 @@ export const SELF_ISSUED_VTC_SERVICE_DESCRIPTION =
 export const SELF_ISSUED_VTC_SERVICE_MINIMUMAGEREQUIRED = Number(
   process.env.SELF_ISSUED_VTC_SERVICE_MINIMUMAGEREQUIRED ?? 18,
 )
+export const SELF_ISSUED_VTC_SERVICE_LOGOURI =
```
genaris, line 115: Make sure to update documentation with all relevant SELF_ISSUED_VTC variables that might have changed.

genaris (changes requested):
I think this is fine and looks conform with v4 specification, where we moved to URIs and digest SRIs for integrity directly on the credentials.

My main concern is that Self issued VTCs are supposed to simplify the initial setup of an agent, and these SRI calculation require the resources to actually exist and be reachable. It is not completely the case right now, because the logo depends on the AGENT_INVITATION_IMAGE_URL, which is optional. Also, we now require SRI for terms and conditions and privacy policy, and if those do not exist, SRI calculation will fail.

We can still keep these variables optional (which is ideal in terms of set up), but we must make sure that agent will start properly. Two ideas that come to my mind:

1) Fallback to a default image for logo, and minimal HTML resource for T&C and privacy policy. This implies that, when these variables are not defined, agent must publish them under its PUBLIC_API_BASE_URL
2) Just set fake values in both. Very simple, but we'll need to check if this would represent an issue on agents that are resolving this credentials (ideally, they will just show the warning icon once they determine that the VTCs are self issued, ignoring the wrong SRI and images)

If it's not so hard to implement, I'd go for the first one. Makes the life easier for users and got a fancy logo right from the start (make sure that the logo you use is from Verana and not Hologram, like the current one).

genaris (approved):

PR #527, opened 2026-07-28
feat(vs-agent)!: enforce required v4 env config

tarunvadde:
Closes #464

- Require `VERANA_CORPORATION_ID`, `VERANA_RPC_ENDPOINT_URL`, `VERANA_INDEXER_BASE_URL` and `ADMIN_API_AUTH_MODE` at startup
- Default `AGENT_DIDCOMM_VERSIONS` to `v1,v2` (v2 is now mandatory)
- Drop the warn-only fallbacks these rules made dead, update the README env tables

Breaking: existing deployments must set the four variables above. The helm chart has no first-class values for the Verana vars yet (use `extraEnv`), the example compose files don't set them either.


apps/vs-agent/src/config/constants.ts
```diff
@@ -30,7 +30,7 @@ export const DEFAULT_PUBLIC_API_BASE_URL = 'http://localhost:3001'
 
 export const ADMIN_API_PUBLIC_URL = process.env.ADMIN_API_PUBLIC_URL
 export const ADMIN_API_EXTERNAL_PORT = Number(process.env.ADMIN_API_EXTERNAL_PORT || 3010)
-export const ADMIN_API_AUTH_MODE = (process.env.ADMIN_API_AUTH_MODE ?? 'internal')
+export const ADMIN_API_AUTH_MODE = (process.env.ADMIN_API_AUTH_MODE ?? '')
```
genaris, line 33: I didn't get this one: ADMIN_API_AUTH_MODE is marked as required, but here we are defaulting to 'none'. Shouldn't we enforce users to set it rather than silently disabling access?

tarunvadde: It's enforced, just not visible in this diff: the empty-list check already existed in main.ts ("ADMIN_API_AUTH_MODE is required (comma-separated list of: internal, corporation)") but was unreachable while the constant defaulted to 'internal'. Removing the default is what makes that check fire when the var is unset.

genaris (approved):

PR #536, opened 2026-07-31
fix(openid4vc): dev-mode robustness - self-heal stale signing records, retry failed initialization

mjfelis:
## What we saw on the playground cast

After the `.5` rollout, the two verifiers with pre-existing wallets (`demo-verifier-accredited`/`-unaccredited`) could not mint OID4VP requests: every attempt failed with

```
KeyManagementKeyNotFoundError: Key with key id 'k3Vhbc7WS2ZNt83ZsLiPHs3Zdt2ZBhRNMjSoxVPdyjiW' not found in backend 'askar'
```

where the "missing" key id is the **legacy base58 id** (base58 of the leaf P-256 compressed public key) of the service's own development certificate - Askar stores keys by uuid. The two fresh services (`demo-issuer-untrusted`/`demo-verifier-untrusted`, first boot ever) were unaffected. **A pod restart healed both broken verifiers**: the failure is a boot-time condition on wallets with prior state, and once it happens the process never recovers on its own because `ensureInitialized()` caches the rejected promise - one verifier kept replaying its original boot error on every request for an hour, which is what made a transient condition look like a permanently corrupt record.

## Fixes

**1. Failed initialization retries (`IssuerService`/`VerifierService`).** A rejected `initialization` promise is no longer cached: reset on rejection so the next request re-runs `initialize()`. One unlucky boot no longer wedges the process until restart.

**2. Stale development signing records self-heal (`CertificateService`).** Development certificates are disposable. If the persisted record is genuinely unusable - expired certificate, identity drift after a DID/host change, malformed record, cert/key mismatch, or a key id no KMS backend holds (records written by older builds) - delete the record and fall through to the existing creation path (which republishes the dev key to the DID document with the correct kid). KMS backend *faults* (`KeyManagementError` other than key-not-found) still propagate: recreating on a transient outage would rotate a healthy certificate and silently break the fingerprints verifiers pinned at deploy time.

## Tests

- failed first init → second `ensureInitialized()` retries and succeeds (issuer + verifier)
- expired persisted dev record → replaced (was: rejected; wedging was the old asserted behavior)
- persisted dev record with a key no backend holds → replaced
- KMS backend fault → propagated, record untouched, no rotation

207/207 plugin tests pass, `tsc` clean.

### DIFF
--- a/packages/plugin-openid4vc/src/services/CertificateService.ts
+++ b/packages/plugin-openid4vc/src/services/CertificateService.ts
@@ -239,23 +239,36 @@ async function loadDevelopmentSigningCertificate(
   const recordId = developmentRecordId(agent.did, hostname, development.commonName, role)
   const existing = await agent.genericRecords.findById(recordId)
   if (existing) {
-    const stored = parseDevelopmentRecord(existing.content)
-    const certificate = X509Certificate.fromEncodedCertificate(stored.certificate)
-    assertCertificateChainUsable([certificate])
-    assertDevelopmentCertificateIdentity(certificate, agent.did, hostname)
+    try {
+      const stored = parseDevelopmentRecord(existing.content)
+      const certificate = X509Certificate.fromEncodedCertificate(stored.certificate)
+      assertCertificateChainUsable([certificate])
+      assertDevelopmentCertificateIdentity(certificate, agent.did, hostname)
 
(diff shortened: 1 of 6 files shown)


genaris (approved):
All this key management might need a refactoring when integrating oidc4vc in main, but if it works for you in the playground, I'm fine to merge here.

PR #541, opened 2026-08-01
fix(model): map v4 networks to their v4 indexer schema routes

mjfelis:
**Changes**
- `mapToEcosystem` still pointed every network at v3 `/verana/*` REST paths. v4 indexers (devnet today) drop those paths and serve `/v4/*` routes on the same host, so devnet schema refs resolved to dead URLs (`https://idx.devnet.verana.network/verana/cs/v1/js/{id}` → 404) and ECS credential classification (`resolveVTCType` / `resolveJSCType`) failed for any credential anchored in a v4 network — the public UI then falls back to the classic view.
- The network map now carries the API version per network: v4 schema refs (canonical `vpr:…:cs:{id}` and expanded `…/cs/v1/js/{id}`) resolve to `{base}/v4/credential-schema/js/{id}`; other v4 paths map onto `{base}/v4`; v3 networks (mainnet placeholder, testnet) keep the `/verana/cs/v1/js` behavior unchanged.
- Added unit tests for `mapToEcosystem` covering canonical/expanded refs on v3 and v4 networks, unknown networks and non-vpr URLs.

**Testing**
- `npx vitest run packages/model/tests/ecs.test.ts` — 7 tests pass.
- `pnpm build` in `packages/model` passes.
- Verified live that `https://idx.devnet.verana.network/v4/credential-schema/js/1` returns the schema JSON while the old `/verana/cs/v1/js/1` path 404s, and that the testnet URLs are unchanged.

## Checklist
- [x] I have added tests to confirm that my fix is effective or that my feature works as intended.

### DIFF
--- a/packages/model/src/utils/ecs.ts
+++ b/packages/model/src/utils/ecs.ts
@@ -5,19 +5,25 @@ export enum ECS {
   USER_AGENT = 'ecs-user-agent',
 }
 
-const urlMap = new Map<string, string>([
-  ['vpr:verana:vna-mainnet-1', 'https://idx.testnet.verana.network/verana'],
-  ['vpr:verana:vna-testnet-1', 'https://idx.testnet.verana.network/verana'],
-  ['vpr:verana:vna-devnet-1', 'https://idx.devnet.verana.network/verana'],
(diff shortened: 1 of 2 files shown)


genaris (commented):
Why would we want to support v3?

PR #542, opened 2026-08-02
feat: ship the openid4vc plugin in the standard vs-agent image

mjfelis:
**Changes**
- Drop the dedicated `vs-agent-openid4vc` image: the size difference against the base image is too small to justify two containers. The openid4vc plugin package is now **built into the standard `vs-agent` image**, where it stays **disabled by default** (`ENV VS_AGENT_PLUGINS=messaging,chat`).
- Enabling stays a pure runtime decision, same mechanics as before: add `openid4vc` to `VS_AGENT_PLUGINS` and provide `OID4VC_CONFIG_FILE` (via the chart: `oid4vc.plugins` + `oid4vc.config`). Nothing changes for deployments that don't use it.
- Dockerfile: the former `build-openid4vc` stage becomes the `vs-agent` build stage (keeping `pnpm install --frozen-lockfile`; the old plugin-less build stage needed `--no-frozen-lockfile` and is gone), the `vs-agent` runtime stage now copies the plugin package, and both `vs-agent-openid4vc` stages are removed. `vs-agent-mrtd` is untouched.
- CI and CD matrices drop the `vs-agent-openid4vc` target; READMEs (root, app, plugin) and the chart's `oid4vc` values comment are updated to describe the enable-at-runtime flow.

**Testing**
- `docker build --check -f apps/vs-agent/Dockerfile .` passes; no references to `vs-agent-openid4vc` or `build-openid4vc` remain outside the changelog.
- CI's docker job builds the remaining targets on this PR.

### DIFF
--- a/apps/vs-agent/README.md
+++ b/apps/vs-agent/README.md
@@ -228,7 +228,7 @@ VS Agent uses an opt-in plugin architecture. Each plugin is an independent packa
 | `messaging` | _(built-in)_                             | Base credential and proof handlers. Always loaded — cannot be disabled.                                                                                    |
 | `chat`      | `@verana-labs/vs-agent-plugin-chat`      | Chat protocols: text messages, media, reactions, receipts, calls, action menus, user profile.                                                              |
 | `mrtd`      | `@verana-labs/vs-agent-plugin-mrtd`      | eMRTD / ePassport verification. Requires the `vs-agent-mrtd` Docker image.                                                                                 |
-| `openid4vc` | `@verana-labs/vs-agent-plugin-openid4vc` | `dc+sd-jwt` issuance and presentation. Requires the `vs-agent-openid4vc` Docker image and [JSON configuration](../../packages/plugin-openid4vc/README.md). |
+| `openid4vc` | `@verana-labs/vs-agent-plugin-openid4vc` | `dc+sd-jwt` issuance and presentation. Ships in the standard `vs-agent` image (disabled by default) and requires [JSON configuration](../../packages/plugin-openid4vc/README.md). |
 
 ### Selecting plugins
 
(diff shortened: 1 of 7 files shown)


apps/vs-agent/README.md
```diff
@@ -290,11 +290,10 @@ First
...
 
-| Target               | Image                  | Plugins included             |
-| -------------------- | ---------------------- | ---------------------------- |
-| `vs-agent`           | `2060io/vs-agent`      | messaging + chat             |
-| `vs-agent-mrtd`      | `2060io/vs-agent-mrtd` | messaging + chat + mrtd      |
-| `vs-agent-openid4vc` | local build target     | messaging + chat + openid4vc |
+| Target          | Image                  | Plugins included                                                    |
+| --------------- | ---------------------- | ------------------------------------------------------------------- |
+| `vs-agent`      | `2060io/vs-agent`      | messaging + chat (openid4vc shipped, enable via `VS_AGENT_PLUGINS`) |
+| `vs-agent-mrtd` | `2060io/vs-agent-mrtd` | messaging + chat + mrtd                                             |
```
genaris: I think this is outdated.


```suggestion
| `vs-agent`      | `veranalabs/vs-agent`      | messaging + chat (openid4vc shipped, enable via `VS_AGENT_PLUGINS`) |
| `vs-agent-mrtd` | `veranalabs/vs-agent-mrtd` | messaging + chat + mrtd                                             |
```

genaris (approved):
I think it wouldn't hurt to have oidc4vc plug-in enabled by default for the sake of this proof of concept with the playground.

And also it could be enabled when integrating in v2.0.0. We just need to make sure that agent configuration mechanisms is consistent among the different plug-ins and modules (e.g. config files, env vars, configSets, etc.).

apps/vs-agent/README.md
mjfelis: Good catch — applied in 3fa7cce (the images publish under `veranalabs/`, the `2060io/` prefix was the stale part).

PR #561, opened 2026-08-05
feat: accept presentations from any accredited issuer of a JSC

lotharking:
**Changes**

* Include issuer and verifier checks against the indexer to validate the state.
* Refactor:
  * Schema reference creation using centralized functions.
  * Indexer creation and inclusion in the agent so it can be reused across all agents.


packages/nestjs-client/src/connections/connection.service.ts
```diff
@@ -28,14 +28,15 @@ export class ConnectionsEventService {
           await this.repository.updateMetadata(event.connectionId, event.metadata)
         await this.handleNewConnection(event.connectionId)
         break
-      case ExtendedDidExchangeState.Completed:
+      case ExtendedDidExchangeState.Completed: {
```
genaris: Why this change?

packages/client/src/services/MessageService.ts
```diff
@@ -47,7 +47,7 @@ export class MessageService {
       let jsonResponse
       try {
         jsonResponse = JSON.parse(responseText)
-      } catch (e) {
+      } catch (_e) {
```
genaris: Doesn't seem needed, is it?

packages/agent-sdk/src/utils/util.ts
```diff
@@ -163,3 +163,18 @@ export async function fetchJson<T>(
     if (timer) clearTimeout(timer)
   }
 }
+
+// TODO: Once all schema refs are migrated to the v4 `:cs:` format, remove the legacy
+// `/cs/v1/js/` protection here (buildLegacySchemaRef and the legacy branch in parseSchemaRef).
```
genaris: What do you mean by "once all schema refs are migrated"? Newer VS Agents would not support v3 (i.e. they won't work on testnet).

Is it anticipating some cases where we want to migrate from vs-agent v1.12.0 to v2.0.0? If so, which ones?

packages/agent-sdk/src/transports/VsAgentWsInboundTransport.ts
```diff
@@ -113,7 +113,7 @@ export class VsAgentWsInboundTransport implements DidCommInboundTransport {
       try {
         const messageReceiver = agentContext.dependencyManager.resolve(DidCommMessageReceiver)
         await messageReceiver.receiveMessage(JSON.parse(event.data), { session })
-      } catch (error) {
+      } catch (_error) {
```
genaris: Also I don't see why this change is introduced in this PR

apps/vs-agent/src/controllers/admin/verifiable/TrustService.ts
```diff
@@ -198,12 +198,13 @@ export class TrustService {
       validateSchema(parsedSchema, claims)
 
       switch (format) {
-        case 'jsonld':
+        case 'jsonld': {
```
genaris: I don't understand this {} rule. It's because it is missing the break? I prefer to keep the same coding style over the codebase

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -203,19 +199,13 @@ export class VtFlowsService {
...
-      if (!VERANA_INDEXER_BASE_URL) {
-        throw new BadRequestException(
-          'Indexer not configured (set VERANA_INDEXER_BASE_URL); required for vt-flow',
-        )
-      }
-      this.indexerService = new VeranaIndexerService({
-        baseUrl: VERANA_INDEXER_BASE_URL,
-        logger: new TsLogger(ADMIN_LOG_LEVEL, 'VeranaIndexer'),
-      })
+  private requireIndexer(agent: VsAgent): VeranaIndexerService {
```
genaris: VERANA_INDEXER_BASE_URL is now mandatory, so I think `indexer` object should not be optional anymore. If so, all this logic would not be required.

apps/vs-agent/src/controllers/admin/invitation/InvitationController.ts
```diff
@@ -254,7 +254,14 @@ export class InvitationController {
...
       schema = schemaResult.schema
-      restrictions = [{ schema_id: schemaResult.schemaId }]
+
+      const jsonSchemaRef = extractJsonSchemaRef(jscData)
+      const accreditedIssuerRestrictions = jsonSchemaRef
+        ? await this.credentialTypesService.resolveAccreditedIssuerRestrictions(jsonSchemaRef)
+        : []
+      restrictions = accreditedIssuerRestrictions.length
+        ? accreditedIssuerRestrictions
+        : [{ schema_id: schemaResult.schemaId }]
```
genaris: This won't work, because the filter will be only by issuer_ids, so ANY credential issued by those issuers will be valid. If you wanted to be very specific, you should add all these issuer_ids AND the schema_id restriction. 

Even if you do so, it is not practical, because for instance you may have 100 allowed issuers, and if their DID is around 50 characters long (to say the least), the resulting DIDComm message will grow in more than 5KB. Although technically possible, we want to keep DIDComm messages as small and predictable as possible.

That's why the current implementation makes use of schema_id as a single filter: it is up to the prover and the verifier to check if the issuer of the credentials they present/receive is authorized. In current vs-agent's code, for example, this check might be done in `BaseMessageEvents.ts`, when handling `DidCommPresentationV2Message.type.messageTypeUri`: based on the credential definition id of the presented credential, we can tell who the issuer is, and right there do the `listParticipants` call to see if it is authorized.

genaris (changes requested):
As discussed in yesterday's WG call, this doesn't seem to be the intended approach to address #386. I think we can proceed on this PR just to get some temporary solution (see my comments), but it will not be actually completed until we properly document the procedure, so both parties (prover and verifier) can agree on how to deal with Proof Exchange flows.

packages/agent-sdk/src/vtFlow/VtFlowOrchestrator.ts
```diff
@@ -398,13 +406,6 @@ export class VtFlowOrchestrator {
     return credentialJson
   }
 
-  private requireIndexer(): VeranaIndexerService {
```
tarunvadde, line 415: Careful on the rebase here, git will not flag this one. #565 added two new `this.requireIndexer()` calls to this file, in `digestAlgorithmForSchema()` and `onCredentialIssued()`. Those hunks are far enough from yours that the merge is clean, so you end up with calls to a method you deleted. I checked with `git merge-tree` and they land at lines 297 and 314 in the merged file.

tsc catches it, so nothing ships broken. The trap is the fix. Restoring `requireIndexer` from main's side compiles and then throws at runtime, because after your change nothing populates `options.indexer` any more. Both call sites need to become `this.agent.indexer`.

Your direction is the right one, `VtFlowsService.ts:170` and `stateMutations.ts:323` already build orchestrators with no indexer on main, so that path is broken today. It just needs finishing.

packages/agent-sdk/src/events/BaseMessageEvents.ts
```diff
@@ -106,6 +108,51 @@ export const baseMessageEvents = async (agent: VsAgent<BaseAgentModules>, logger
...
+                ref: callbackParameters.ref,
+              }),
+            )
+          }
+
+          const trustRegistry = record.metadata.get('_2060/trustRegistry') as
+            | { schemaId?: number }
+            | undefined
+
+          if (trustRegistry?.schemaId !== undefined) {
```
tarunvadde, line 142: This runs after Credo has already accepted and acked the presentation, and the rejection is never persisted.

I think by the time this line runs the record is `state: done, isVerified: true`. `sendProblemReport` only builds and sends the message, it does not `updateState`, unlike `declineRequest` in the same API.

So `GET /presentations/:id` reports `verified: true` for a presentation we explicitly rejected as untrusted.

I agree with genaris that this file is the right place for the check, but a post-dispatch hook is too late to enforce anything. Needs `shouldAutoRespondToPresentation` or `autoAcceptProof: Never` on these requests, plus `updateState(Abandoned)` with an errorMessage so the record carries the decision.

packages/agent-sdk/src/events/BaseMessageEvents.ts
tarunvadde, line 237: If the indexer is down this whole check just goes quiet. `findUnaccreditedDids` runs `Promise.all` over `listParticipants`, and `fetchJson` throws on any non-2xx and on the timeout. That lands in the try above whose only handler is a `logger.error`.

So you get no `UNTRUSTED_ISSUER`, no `OK`, no error, one log line, and the holder already has a success ack. Nothing for the consumer to poll on.

The odd part is the asymmetry. An indexer that says "no participants" fails closed. An indexer that doesn't answer says nothing at all. A check that can't run should fail the exchange.

Also worth knowing, since it's `Promise.all`, one unreachable DID kills the check for every credential in the presentation.

tarunvadde (changes requested):

tarunvadde (approved):
LGTM. Please resolve conflicts.

genaris (changes requested):
I suggest to hold this PR until we refactor several things on VS Agent API for v2.0.0. One of these changes is that we are going to make relatedJsonCredentialSchemaId mandatory for every time of credential flow, so several things on the code will be affected and change.

genaris (commented):
I reviewed branch `b4073739` against #386 and against the AnonCreds rules of Verifiable Trust v4 (CIB-3, CIT-3, PRB-3, PRT-3, TR-5). The VS Agent spec now defines the full procedure in [VSA-VTI-FLOW-VERIFY-AC](https://github.com/verana-labs/verana-spec/blob/main/v4/vs-agent/spec.md#vsa-vti-flow-verify-ac-anoncreds-trust-decision). This review refers to those identifiers.

Result: the verifier check has the correct shape. The PR does not solve #386 for a second issuer. 3 of the 6 required checks are missing.

## Blocking problems

1. **A second issuer gets a private AnonCreds schema.** `getOrRegisterAnonCredsSchema` (`CredentialTypeService.ts:335-343`) registers a new schema under the issuer DID if no local record exists. The issuance paths (`BaseMessageHandler.ts:199`, `TrustService.ts:272`) call it without `issuerDid`. The verifier restricts the request by the Ecosystem `schema_id`. The credentials of that issuer never match. The e2e test does not show this, because its only issuer is the VTJSC issuer. Spec: VSA-PUB-AC-5, VSA-ADM-AC-CD-CREATE. Issue #676  (part 1).

2. **The issuer self-check does not run for a credential definition on a remote schema.** `resolveTrustRegistrySchemaIdForCredentialDefinition` (`CredentialTypeService.ts:41-58`) reads the VTJSC id from the local schema record only. For a remote schema it returns `undefined`. `assertAccreditedForSchema` then returns without a check (lines 60-63). It also returns if the agent has no DID. Spec AC-1: a check with missing inputs fails. It does not skip. Issue #677 (part 2).

3. **The holder check and the prover check do not exist.** CIT-3 and PRT-3 require them. Nothing runs on `offer-received` or on `request-received`. Spec AC-4 and AC-6. Issue #679 (part 4).

4. **The verifier self-check runs on 1 of 3 request paths.** Only the `jsonSchemaCredentialId` path of the invitation endpoint checks (`V1InvitationController.ts:269`). The `credentialDefinitionId` path (line 291) and the identity-proof-request (`BaseMessageHandler.ts:342`) do not check. They store no schema id. As a result, the presentation check does not run for them. Spec AC-5. Issue #678 (part 3).

## Defects in the presentation handler

- The handler sets `record.state = Abandoned` and calls `proofs.update` (`BaseMessageEvents.ts:164-165`). This does not call `updateState`. No `DidCommProofStateChanged` event occurs. On main, the backend gets no `didcomm.presentations.state-updated` event. `isVerified` stays `true`. `sendProblemReport` does not change the state. Spec AC-7. Issue #680 (part 5).
- The handler reads `$ref` from the VTJSC without a proof check and without an Ecosystem DID check (TR-3). It does not compare the `relatedJsonSchemaCredentialId` of the credential definition with the requested VTJSC. Spec AC-1 and AC-7. Issues #676 and #680 (parts 2 and 5).
- The handler abandons the exchange if the indexer is not available. This is correct (AC-8). Keep it.

## Other points

- Names. Use CredentialSchema and Ecosystem. Do not use "trust registry" (`resolveTrustRegistrySchemaId`, `_2060/trustRegistry`, `e.p.trust-registry-unavailable`). The spec codes are `e.p.issuer-not-authorized`, `e.p.trust-resolution-unavailable`, `NOT_AUTHORIZED` (409), `PEER_NOT_AUTHORIZED` (409) and `RESOLVER_UNAVAILABLE` (503).
- A request for a VTJSC without a `vpr:` reference fails with 400. An offer for the same VTJSC skips the check. Make both fail.
- `parseSchemaRef` accepts the legacy `/cs/v1/js/` form. `mapToEcosystem` rejects it. Use one form.
- The `validate` script change and the indexer plumbing (`main.ts`, `EcsBootstrapService`, test mocks) are not related. Merge them separately.
- The branch conflicts with main in 5 files. The v2 controllers on main have none of these checks. Put the checks in one shared service, and wire it into the v2 controllers only. The v1 controllers get no change.

## Keep

- `VeranaIndexerService.findUnaccreditedDids` and its unit tests.
- The e2e harness in `accreditedIssuers.e2e.test.ts`.
- `autoAcceptProof: Never` with an acknowledgement after the check.

## Next step

I close this PR. Issues #676 #677, #678 #679  and #680 replace it. If you want to keep the branch, rebase it on main and target it.

PR #565, opened 2026-08-06
feat(vt-flow)!: anchor the signed credential digest via verre

tarunvadde:
Closes #562, #563, #564.

One PR because the three are not separable. verre's `computeCredentialDigestJCS` returns a bare digest, so adopting verre (#563) is the prefix fix (#562), there is no separate code for it. And once `proof` is inside the hash, the digest can only be taken after signing, which is #564. Doing #563 first would break the roundtrip, the issuer hashing unsigned while the holder hashes signed. Dropping the proof-stripping intermediate removed the only split point.

- The digest used to be computed from the unsigned credential and anchored at offer time. Now the validator signs first, holds the message unsent, digests it, anchors it, and only then delivers. A failed anchor aborts the issuance and terminates the session with a problem report instead of shipping a credential nobody can verify.

- The digest comes from verre's `computeCredentialDigestJCS`, so it covers the whole credential including `proof`, it has no algorithm prefix, and the algorithm is read from the schema's `digest_algorithm` rather than hardcoded. The local `credentialDigest.ts` is gone, vs-agent no longer computes a credential digest anywhere.

- Also fixes `op_summary_digest`, which was being sent the credential digest. Per MOD-PP-MSG-3 that field digests the applicant's submission, so it is now omitted.

Two things to know before merging:

- devnet is still on v0.10.1-dev.25, which rejects an unprefixed digest at CheckTx. https://github.com/verana-labs/verana-node/pull/27 is merged but not deployed.
- once https://github.com/verana-labs/verana-indexer/pull/383 lands, credentials anchored with the old `sha384-` digest can never match and will need re anchoring.


packages/agent-sdk/tests/e2e/helpers.ts
```diff
@@ -14,7 +14,7 @@ const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:v0
 const VERANA_PLATFORM = process.env.FLOW_VERANA_PLATFORM || 'linux/amd64'
 const POSTGRES_IMAGE = process.env.FLOW_POSTGRES_IMAGE || 'postgres:16-alpine'
 const REDIS_IMAGE = process.env.FLOW_REDIS_IMAGE || 'redis:7-alpine'
-const INDEXER_IMAGE = process.env.FLOW_INDEXER_IMAGE || 'veranalabs/verana-indexer:dev'
+const INDEXER_IMAGE = process.env.FLOW_INDEXER_IMAGE || 'veranalabs/verana-indexer:v2.0.0-dev.74'
```
genaris: Why do we need to tag the exact version here?

packages/vt-flow/src/VtFlowApi.ts
```diff
@@ -255,6 +270,84 @@ export class VtFlowApi {
...
+    connectionRecord?.assertReady()
+
+    const requestMessage = await protocol.findRequestMessage(this.agentContext, credentialExchangeRecord.id)
+    if (!requestMessage) {
+      throw new CredoError(`No request message found for credential record '${credentialExchangeRecord.id}'`)
+    }
+    const offerMessage = await protocol.findOfferMessage(this.agentContext, credentialExchangeRecord.id)
+    if (!offerMessage) {
+      throw new CredoError(`No offer message found for credential record '${credentialExchangeRecord.id}'`)
+    }
```
genaris: I'm fine on leaving this, but do we really need them? This was used for connection-less flows in Credo, not sure if we need them under DIDComm V2. But as I said, no problem if they are still there.

genaris (approved):
LGTM! Just some minor comments. Good to merge.

packages/vt-flow/src/VtFlowApi.ts
tarunvadde: Oh yes, this is basically dead code I should've checked I copy-pasted directly from the reference implementation here https://github.com/verana-labs/vs-agent/issues/564. Thanks for pointing out. Removed.

packages/agent-sdk/tests/e2e/helpers.ts
tarunvadde: No we don't need it. Pushed it accidentally. Reverted to :dev.

PR #572, opened 2026-08-07
fix(agent-sdk): pin the e2e chain to dev.30 for the operator grant convention

tarunvadde:
#565 changed `grantOperatorAuthorization` to send `operator: policyAddress`. That is correct, but only from verana-node v0.10.1-dev.30.

### DIFF
--- a/packages/agent-sdk/tests/e2e/helpers.ts
+++ b/packages/agent-sdk/tests/e2e/helpers.ts
@@ -9,8 +9,8 @@ import type {
 import { GenericContainer, Network, Wait, type StartedTestContainer } from 'testcontainers'
 import WebSocket from 'ws'
 
-// applicant-ops needs dev.21+ (VSOA accepted for SetParticipantOPToValidated).
-const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:v0.10.1-dev.21'
+// v0.10.1+: MsgGrantOperatorAuthorization only skips the delegation check when operator == corporation.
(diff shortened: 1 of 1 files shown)


packages/agent-sdk/tests/e2e/helpers.ts
```diff
@@ -9,8 +9,8 @@ import type {
 import { GenericContainer, Network, Wait, type StartedTestContainer } from 'testcontainers'
 import WebSocket from 'ws'
 
-// applicant-ops needs dev.21+ (VSOA accepted for SetParticipantOPToValidated).
-const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:v0.10.1-dev.21'
+// dev.30+: MsgGrantOperatorAuthorization only skips the delegation check when operator == corporation.
+const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:v0.10.1-dev.30'
```
genaris:
```suggestion
const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:v0.10.1'
```
or
```suggestion
const VERANA_IMAGE = process.env.FLOW_VERANA_IMAGE || 'veranalabs/verana-node:dev'
```

Should work

genaris (approved):

packages/agent-sdk/tests/e2e/helpers.ts
tarunvadde: But verana-node doesn't have the :dev tag, right? I couldn't find it if there is.

tarunvadde: I took `v0.10.1`, thanks.

PR #573, opened 2026-08-07
feat(agent-sdk): serve the ECS badge schema and 404 unknown schemaIds

tarunvadde:
Closes #569.

Adds the published v4 badge schema to `getEcsSchemas` and wires it into `ECS_TITLE_BY_TYPE` and the UI's `TYPE_ORDER`. The reference digest in `ecs.ts` was already correct, only the schema itself was missing, so all five ECS types now round-trip through `identifySchema`.

`SelfTrController.getSchema` returned the raw map lookup, so an unknown schemaId gave 200 with an empty body. It now throws 404. The old `try/catch` also swallowed its own deliberate 404 and rethrew it as a 500, so that guard never worked. Removed, since the body is an in memory lookup that cannot throw.

### DIFF
--- a/apps/vs-agent/src/controllers/public/self-tr/SelfTrController.ts
+++ b/apps/vs-agent/src/controllers/public/self-tr/SelfTrController.ts
@@ -46,15 +46,11 @@ export class SelfTrController {
   @ApiParam({ name: 'schemaId', required: true, description: 'Schema identifier', example: 'ecs-org' })
   @ApiResponse({ status: 200, description: 'JSON schema returned' })
   async getSchema(@Param('schemaId') schemaId: string) {
-    try {
-      if (!schemaId) {
-        throw new HttpException('Schema not found', HttpStatus.NOT_FOUND)
-      }
-      return this.ecsSchemas[schemaId]
-    } catch (error) {
-      this.logger.error(`Error loading schema file: ${error.message}`)
-      throw new HttpException('Failed to load schema', HttpStatus.INTERNAL_SERVER_ERROR)
+    const schema = this.ecsSchemas[schemaId]
+    if (!schema) {
+      throw new HttpException('Schema not found', HttpStatus.NOT_FOUND)
     }
+    return schema
   }
 
   @Get('perm/v1/list')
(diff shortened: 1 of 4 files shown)


apps/vs-agent/src/controllers/public/self-tr/SelfTrController.ts
```diff
@@ -46,15 +46,11 @@ export class SelfTrController {
...
-      this.logger.error(`Error loading schema file: ${error.message}`)
-      throw new HttpException('Failed to load schema', HttpStatus.INTERNAL_SERVER_ERROR)
+    const schema = this.ecsSchemas[schemaId]
+    if (!schema) {
+      throw new HttpException('Schema not found', HttpStatus.NOT_FOUND)
     }
+    return schema
   }
 
   @Get('perm/v1/list')
```
genaris: This outdated paths makes me think that this could represent an issue when resolving these self-issued credentials. Can you check how are the indexer's trust resolver module/verre processing these credentials and verify if it would work as expected?

If these API methods are not used, I prefer to delete them rather than having some dead code.

genaris (commented):
OK with the changes done here but please verify what's going on with that outdated perm/list.

apps/vs-agent/src/controllers/public/self-tr/SelfTrController.ts
tarunvadde: Removed, and you were right to pull on it.

It was a stub, not just an outdated path. It queried nothing and returned a fabricated ISSUER permission for any DID, with a hardcoded created date of 2000-11-18. Introduced in 9b09186, "refactor self permission for testing porpose", which renamed the older perm/v1/find_with_did and reshaped the fake to match the VPR list format.

It made sense under v3, where verre asked the agent whether a DID was an authorized issuer, so an agent had to vouch for its own self-issued credentials. v4 moved that on-chain, verre 0.4.0 resolves participants via the indexer's /v4/participant/list and never calls an agent-hosted permission endpoint.

Confirmed nothing depends on it: zero references in verre source, zero in the verre 0.4.0-dev.1 build we ship, zero in the indexer, zero elsewhere in vs-agent, zero in the v4 spec. Only the route definition itself.

Also dropped Query and ApiQuery, which it was the last user of.

genaris (approved):

Issue #574, opened 2026-08-10
Indexer event stream does not follow the connect-first order from spec v4-draft4


tarunvadde:
Ariel asked us to check if the agent follows the new spec text merged in https://github.com/verana-labs/verana-spec/pull/37 (connect-first order for the indexer event stream).

## What is already correct

- The agent keeps new events aside while it reads the old ones, then handles the old ones first and the new ones after, in the right order.
- It skips events already handled, checking both things the spec asks for, the block number and the event key. `IndexerWebSocketService.ts:224` and `:226`.
- It only saves the position after an event is fully handled, and always saves one block behind. So if it stops in the middle of a block, that block is read again next time. `:239`.
- When the connection drops and comes back, the same steps run again.

## What is wrong

The agent starts reading old events too early.

In `IndexerWebSocketService.ts` the socket opens at line 83, but that call does not wait for anything. At line 86 the agent already starts reading old events. The `subscribe` message is only sent later, when `ready` arrives, at line 149. So both things happen at the same time.

If a new block arrives after the old events are read, but before the indexer marks the client as subscribed, nobody delivers that block. The indexer skips clients that are not subscribed yet (`subscribe_broadcaster.ts:39`) and it never resends old blocks. The read of old events already finished, so it will not come from there either. That event is lost for good.

## Why the agent cannot fix this alone

The indexer never answers the `subscribe` message. In `handleControlMessage` it saves the client state and returns, sending nothing back. So the agent has no way to know when it is really subscribed.

Waiting until the message is sent makes the problem smaller. It does not remove it.

One of these is needed in the indexer:

1. Keep blocks for a client from the moment it connects. The client list is already there.
2. Answer the `subscribe` message, so the agent knows when to start reading old events.

Until then, this line in `verana-indexer/spec.md:1352` and `:1410`, that the socket sends every block from `ready.block`, is not true. The same happens on the VT socket (`vt_subscribe_broadcaster.ts:46`).

## Smaller thing

Live events are not sorted, but old events are (`:177` and `:195`). This is fine while the order coming in is correct. But `persistIndexerEventsForBlock` reads ids in groups of 500 and adds delegation events at the end, so for a block with more than 500 events the order may not be correct.

## What to do

- Agent side: wait for the `subscribe` message to be sent before reading old events. This needs a timeout, because `main.ts` waits for `start()` before DIDComm is ready, so a dead indexer must not block startup.
- Indexer side: needs a decision on option 1 or 2 above.

[...]

PR #575, opened 2026-08-10
fix(agent-sdk): send subscribe before draining indexer catch-up

tarunvadde:
Part of #574.

`connect()` opened the socket and started the REST drain right away, while `subscribe` only went out later when `ready` arrived. An event landing in between was delivered by neither.

Now it waits for the subscribe frame to be sent. A dead socket releases the wait on error or close, and a 5s timeout stops a slow indexer blocking startup.

Does not fully close the gap. The indexer sends no reply to `subscribe`, so we still cannot tell when it actually takes effect. That part is in #574.


genaris (approved):
Ok, better than nothing. Let's follow up this problem on https://github.com/verana-labs/vs-agent/issues/574.

PR #576, opened 2026-08-11
fix(demo): make the v4 lifecycle demo run end to end

genaris:
## Summary

The local demo in `docs/v4-lifecycle.md` could not reach the onboarding flow. Six
separate defects blocked it, and each one hid the next. This PR clears all six.

The demo now runs from a clean start to a live DIDComm flow between the two agents.

One fix lands in `agent-sdk`. Everything else is demo material: the seed, compose,
the chain init script, the env example, and the docs.

## The one product fix

`setupSelfTr` now publishes the `vpr-schemas-<key>-vtc-vp` service and republishes
the DID document.

A resolver discovers a credential only through that fragment, and `#whois` does not
match it. The previous code renamed a service that already carried the schema key,
but no code ever created one, so the branch never ran. It also wrote to the DID
repository alone, which kept the new service local while resolvers read the
published document.

Without this, a Verifiable Service that presents self-issued credentials is
invisible to a resolver, whatever the chain holds.

## Demo topology: three corporations

The seed now creates three corporations, and each one holds a single role.

| Corporation | Id | Holds |
|---|---|---|
| Validator | 1 | the validator's participants |
| Ecosystem | 2 | the ecosystem, the ECS schemas, the root participants |
| Applicant | 3 | the applicant's participants |

An agent publishes a VTJSC for every schema its own corporation owns, and that
credential references the schema on chain. Both agents here present self-issued
credentials, which must reference only their own URLs, so neither may own a schema.
No agent operates for the ecosystem corporation.

The validator and the applicant also need separate corporations. A participant OP
is unique per (schema, role, validator, authority), so one shared corporation makes
the applicant's Service onboarding collide with the validator's.

## Other demo fixes

- **Indexer race.** The indexer had no healthcheck, and both agents waited only for
  `service_started`. They raced ahead and failed with `ECONNREFUSED`. The indexer
  now has a healthcheck, and both agents wait for `service_healthy`.
- **Mnemonic collision.** `.env.demo.example` gave both agents the `cooluser`
  mnemonic, which the seed itself signs with. The seed's operator grant therefore
  collided with the validator's, and the seed always failed. Each agent now has its
  own throwaway key.
- **Applicant grant.** The seed never funded the applicant operator and never gave
  it an OperatorAuthorization, so the applicant could never pass its bootstrap
  preflight check. The seed now does both, through the new
  `DEMO_APPLICANT_OPERATOR` variable.
- **Real ECS schemas.** The seed registered placeholder schemas. `identifySchema`
  matches an ECS schema by its SHA-384 digest, so the agents never recognised them.
  The seed now registers the published v4 schemas through `getEcsSchemas`.
- **Validator Service participant.** The validator needs an active ISSUER
  participant on the Service schema, or the applicant's VS-CONN-VS check fails.
- **Chain id.** The demo chain used `vna-testnet-1`, and `mapToEcosystem` maps that
  id to `https://idx.testnet.verana.network`. A local reference could therefore
  reach a public host and return the wrong schema without an error. The demo chain
  is now `vna-demo-1`.
- **Node image.** The compose default moves from `v0.10.1-dev.25` to `v0.10.1`.
- **`VeranaTestChain.grantOperatorAuthorization`** declared a `grantee` parameter
  and then ignored it, so it could grant only to the caller.

## Documentation

`docs/v4-lifecycle.md` now describes the three corporations, collects both operator
addresses, and uses `restart` rather than `up -d`. Compose recreates a container
only when its configuration changes, and `.env` already holds the seeded values, so
`up -d` restarts nothing.

## How to verify

```bash
cp .env.demo.example .env
docker compose -f docker-compose.demo.yml --env-file .env up --build -d
# collect both vs_operator addresses, then seed, then:
docker compose -f docker-compose.demo.yml --env-file .env restart agent-validator agent-applicant
```

Expected result: `applicant OR_SENT` and `validator AWAITING_OR` on one session id.
The validator's credential chain stays local:

```
credentialSchema : https://agent-validator.demo/vt/schemas-example-service-jsc.json
jsonSchema.$ref  : https://agent-validator.demo/vt/cs/v1/js/ecs-service
```

## What is left

**Review this one first.** A self-issued credential returns after a restart. When a
real credential arrives, `updateVtcEntries` detaches the self-issued entries. The
early return in `generateVerifiablePresentation` then fails on the next start,
because it requires `attached`, so the agent rebuilds the credential and writes
`attached: true`. This PR makes that visible in the published DID document, because
the agent now republishes the service. A resolver takes the first service credential
it finds, so a self-issued credential could outrank a real one. A fix is to keep a
detached entry detached, and to restore it only through an explicit removal.

Other items, none of which block the demo:

- `publishVtjscIfOwner` performs no ownership check, although its name promises one.
  The `CreateNewCredentialSchema` handler calls it for every schema on the chain.
- `GET /vt/cs/v1/js/<unknown>` answers 200 with an empty body. `getSchema` raises a
  404, so a handler ahead of the controller answers that path.
- The testcontainers stack in `packages/agent-sdk/tests/e2e/helpers.ts` still uses
  `vna-testnet-1`, so it carries the same public-host hazard the demo just lost.
- `fullLifecycle.e2e.test.ts` stubs `assertVerifiableService: async () => true`, so
  no test exercises the real VS-CONN-VS path. A test there would have caught the
  missing linked presentation service.
- A self-issued service resolves with outcome `NOT_TRUSTED` and `verified: true`.
  `VERIFIED_TEST` needs a `vpr:` reference with a non-production registry, and that
  reference brings back the anchored digest requirement. This deserves a decision.
- The applicant's Service onboarding cannot complete. Its validator is
  `did:example:demo-service-root`, a placeholder with no agent behind it. The
  participant OP exists on chain and stays PENDING.

### DIFF
--- a/.env.demo.example
+++ b/.env.demo.example
@@ -1,17 +1,23 @@
 # Copy to .env and adjust. Used by docker-compose.demo.yml.
 
-# Funded demo account from docker/verana-node-init.sh (cooluser). Replace per agent
-# with distinct funded accounts once you have seeded corporations and grants.
-VALIDATOR_MNEMONIC="pink glory help gown abstract eight nice crazy forward ketchup skill cheese"
-APPLICANT_MNEMONIC="pink glory help gown abstract eight nice crazy forward ketchup skill cheese"
+# Throwaway local-devnet keys, one per agent. These must stay distinct from each other and
+# from the cooluser mnemonic in docker/verana-node-init.sh that the seed script signs with:
+# the seed grants the validator a VSOA and the applicant an OperatorAuthorization on the same
+# corporation, and the chain rejects an account that would end up holding both.
+# The seed funds both accounts, so they need no balance up front.
+VALIDATOR_MNEMONIC="recall buyer young reopen napkin will office warm chronic witness rigid wood"
+APPLICANT_MNEMONIC="oblige wild human tomato announce method slim buzz maid seat chunk raven"
 
-# Set after seeding: the corporation each agent operates for.
-VALIDATOR_CORPORATION_ID=
-APPLICANT_CORPORATION_ID=
+# The corporation each agent operates for. The seed creates three corporations on a fresh chain,
+# in this order: the validator, the ecosystem owner, and the applicant. Neither agent operates for
+# the ecosystem corporation. Both agents require these values at startup, and the seed prints the
+# ids it created so you can confirm them.
+VALIDATOR_CORPORATION_ID=1
+APPLICANT_CORPORATION_ID=3
 
-# Set after seeding: comma-separated DIDs of the ECS ecosystems the agents trust (WL-ECS).
-TRUSTED_ECS_ECOSYSTEM_DIDS=
+# Comma-separated DIDs of the ECS ecosystems the agents trust (WL-ECS). The seed creates this one.
+TRUSTED_ECS_ECOSYSTEM_DIDS=did:example:demo-ecosystem
 
 # Optional image overrides.
-#VERANA_NODE_IMAGE=veranalabs/verana-node:v0.10.1-dev.25
+#VERANA_NODE_IMAGE=veranalabs/verana-node:v0.10.1
 #VERANA_INDEXER_IMAGE=veranalabs/verana-indexer:dev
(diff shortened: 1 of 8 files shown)


tarunvadde (approved):
Does anything in the demo path still resolve schema refs through `mapToEcosystem` in packages/model? It only maps vna-testnet-1 and vna-devnet-1, so a `vpr:verana:vna-demo-1:cs:N` ref falls through unchanged. verre is fine since the registry scheme comes from VERANA_CHAIN_ID, so probably nothing, just checking there is no third resolution path.

PR #585, opened 2026-08-11
fix(agent-sdk): keep self-issued ECS credentials detached once a real one is stored

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/578

Two bugs, both around the `attached` flag on the self-issued ECS entries.

`saveMetadataEntry` detached them on every write, including `_vt/jsc`. So publishing any JSON schema credential stripped the agent's own ECS services from the DID document, even with no real credential involved. Only a real `_vt/vtc` write does that now.

`generateVerifiablePresentation` wrote `attached: true` on every rebuild and its early return required `attached`, so a detached entry came back on the next restart. It now keeps the stored value and only touches the DID document while the entry is attached. That also repairs the restore path, since `updateVtcEntries(true)` no longer sees a flag that already says true.

One test, on the `_vt/jsc` case. I checked the others I had written against main and they passed there, so they were testing existing behavior and I dropped them. The restart half needs the real signing path, so it is not covered here.

### DIFF
--- a/packages/agent-sdk/src/utils/setupSelfTr.ts
+++ b/packages/agent-sdk/src/utils/setupSelfTr.ts
@@ -332,7 +332,8 @@ export async function generateVerifiablePresentation(
   const integrityData = buildIntegrityData({ id, type, credentialSchema, claims })
   const record = didRecord.metadata.get('_vt/vtc') ?? {}
   const metadata = record[credentialSchema.id]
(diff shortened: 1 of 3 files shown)


genaris (approved):
In my tests on devnet I didn't experience this issue but I don't see any problem on adding this.

genaris (approved):

PR #588, opened 2026-08-12
fix: store digest of ECS credentials (self issuing case) 

genaris:
I noticed that when a VS Agent issues an ECS credential to itself because it has proper permissions to do so, it's not anchoring its digest on blockchain, so other agents cannot check for issuance time. This is different from the case of the initial Self signed credentials (only for test purposes, not anchored on VPR).

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -114,7 +114,7 @@ jobs:
 
   docker:
     needs: version
-    if: needs.version.outputs.publish == 'true'
+    if: ${{ !cancelled() && needs.version.outputs.publish == 'true' }}
     runs-on: ubuntu-latest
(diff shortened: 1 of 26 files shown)


tarunvadde (approved):
LGTM. 

One small thing, not blocking. If the anchoring fails, anchorCredentialDigest only logs a warning, so the credential is still published but has no digest on chain. verre will then reject it with NO_ANCHORED_DIGEST, so the problem shows up later and is harder to trace.

In https://github.com/verana-labs/vs-agent/pull/586 I made the same step throw instead, so nothing gets published without an anchor. Worth deciding if both should work the same way.

tarunvadde (approved):
Thanks! Looks much better now.

PR #589, opened 2026-08-12
fix: ECS credential anchoring, cross-Corporation issuers, and four fixes found on devnet

genaris:
## Changes

**1. Self-issued ECS credentials anchor with `CreateOrUpdateParticipantSession`; `MsgStoreDigest` is removed**
`utils/trustCredentialStore.ts`, `VeranaChainService.storeDigest` (deleted).
`StoreDigest` requires an `OperatorAuthorization`. `CreateOrUpdateParticipantSession` requires a
`VSOperatorAuthorization`, and the chain forbids one account from holding both. An agent that
anchored with `StoreDigest` therefore could never also deliver a credential to a third party. CSPS is also what the VT spec asks of an issuer; the chain stores the digest keeper-to-keeper, so the `getDigest` idempotency check is unchanged. A self-issued credential names only the issuer in the session, because it has no counterparty. The caller passes the ISSUER Participant id, which it already has.

**2. Self-issued credential reconciliation follows the agent's ISSUER Participants**
`stateMutations.reconcileSelfIssuedEcsCredentials`.
It previously iterated the Ecosystems of the agent's own Corporation. That assumes an agent only ever issues inside its own Corporation, and that an Ecosystem controller always holds an ISSUER entry — neither is true. Entries are filtered on `vs_operator === chain.address`, since the chain accepts the anchoring CSPS from no other signer. VTJSC *publication* still follows the agent's own Corporation: only an Ecosystem publishes thoffse.

**3. The VTJSC is resolved from the Ecosystem instead of the agent's own record** — new
`utils/vtjscResolver.ts`
`buildCredential` read the VTJSC from the agent's own DID record and used that copy's URL as `credentialSchema.id`, which only works if the issuer republishes the Ecosystem's VTJSCs. An issuer admitted by another Corporation's Ecosystem holds no copy and must not publish one. It now reads the Ecosystem controller's DID Document and follows `#vpr-schemas-<id>-vtjsc-vp`. An agent that controls the Ecosystem answers from its own record and makes no network call.

**4. `ListParticipants` replaces `FindParticipantsWithDID`**
verana-node does not implement that RPC; calls fail with `unknown query path`. It was unreachable in `findActiveIssuerParticipantId` and `findActiveHolderParticipantIdByDid`. The node applies the request filters loosely, so we ask for the maximum page and match the fields client-side.f

**5. Reconciliation is re-triggered by chain events** — new `blockchain/handlers/selfIssuanceHandlers.ts`
Reconciliation ran only at start. An ISSUER Participant created after the agent booted was never
picked up, so its credential stayed unanchored until the next restart. Wraps
`SetParticipantOPToValidated` and `SelfCreateParticipant` — composing with the default handlers, as `registerAuthorizationHandlers` does — checks the participant is this agent's own ISSUER, and
reconciles.

**6. `EcsBootstrapService.acceptPendingOffers()` runs before the preflight gate**
It re-drives an interrupted credential offer and signs nothing on chain, so a skipped preflight
should not skip it too. The skip message now also states that the agent expects its Participants from out of band, which reads as a condition rather than a failure.

**7. The flow admin API passes the indexer to the orchestrator**
`POST /v1/vt/flows/:id/validate` failed with `Agent has no indexer configured` on agents that had
one. `VtFlowsService` already builds one in `getIndexer()`; it never passed it.

**8. `vt-flow` accepts claims in `AWAITING_OR`**
`updateClaims` allowed only VALIDATING and CRED_REVOKED, but `validateAndOfferCredential` requires AWAITING_OR and accepts, validates and offers in one step. A validator had no state in which it could set claims, and the applicant sends none, so the credential was built with an empty subject and rejected by its own schema check.

**9. The gas adjustment is configurable** — `VERANA_GAS_ADJUSTMENT`, default `1.5`
`broadcastMsg` used cosmjs `'auto'`, i.e. simulate × 1.4. A simulation signs with an empty signature
and runs against the state of the moment, so it under-reports; cosmjs says so in its source and
points at cosmos-sdk#16020. On devnet `TriggerResolver` simulated 37602, needed 1.41 and got 1.4,
then failed out of gas. No exact value exists before delivery, which is why every Cosmos tool has
`--gas-adjustment`; 1.5 matches what the operator scripts pass to `veranad`.

## Docs

`apps/vs-agent/README.md` gains `VERANA_GAS_ADJUSTMENT` and states what `VERANA_ACCOUNT_MNEMONIC`
must and must not be.

## Verified on devnet

An Organization credential issued by a Verifiable Service in its own Corporation, to a service in a second Corporation, against the Ecosystem's VTJSC in a third. Both agents resolve `verified=true`.

## Tests

New: `vtjscResolver.test.ts` (4), `selfIssuanceHandlers.test.ts` (5). Updated:
`ecsDigestAnchoring.test.ts` for the CSPS path. Full suite, typecheck and lint pass.

## Follow-up (not in this PR)

- Delete published credentials: an agent cannot drop or rebuild a published ECS credential that stops verifying, and the admin escape hatches sit behind `IndexerDisabledGuard`.
- Avoid using chain Query API directly: in several places, VeranaChain is using blockchain Query API, while the intended way would be to use indexer for queries (leave only the chain usage for transaction creation)
- Updates on vt-flow specification to accept the required states (cc @tarunvadde)


packages/agent-sdk/src/blockchain/VeranaChainService.ts
```diff
@@ -350,21 +350,32 @@ export class VeranaChainService {
 
   async findActiveHolderParticipantIdByDid(did: string): Promise<number | undefined> {
     // fromPartial fills the unused fields with defaults so the request encodes correctly.
-    const request = QueryFindParticipantsWithDIDRequest.fromPartial({ did })
-    const { participants } = await this.ppQuery.FindParticipantsWithDID(request)
+    const request = QueryListParticipantsRequest.fromPartial({
+      did,
+      role: PARTICIPANT_ROLE_HOLDER,
+      responseMaxSize: PARTICIPANT_QUERY_MAX_SIZE,
```
tarunvadde: The pinned verana-types codec (0.10.1-dev.16) only encodes modifiedAfter and responseMaxSize, so the did, role, schemaId and grantee passed here never reach the wire. The node then returns the first 1024 participants of the whole chain sorted by modified ascending, so the most recently modified entries, our own participant included, are the first ones dropped once the chain outgrows a page, and there is no pagination loop. It surfaces as ANCHORING_FAILED for a participant that exists. Bumping verana-types to 0.10.2 gives the real server side filters, or we loop on modifiedAfter.

tarunvadde (approved):

apps/vs-agent/README.md
```diff
@@ -122,14 +122,15 @@ These variables connect the agent to the Verana network (permission management,
...
 | `VERANA_RPC_ENDPOINT_URL`                  | REQUIRED    | Verana blockchain RPC endpoint URL.                                                                                                                                                                                                     |
-| `VERANA_ACCOUNT_MNEMONIC`                  | REQUIRED    | BIP-39 mnemonic for the agent's Verana blockchain account.                                                                                                                                                                              |
+| `VERANA_ACCOUNT_MNEMONIC`                  | REQUIRED    | BIP-39 mnemonic for the agent's Verana account. This account is the agent's `vs_operator`: it holds only the `VSOperatorAuthorization` records granted on the agent's own Participant entries. It MUST NOT be the account that holds the Corporation's `OperatorAuthorization` — the chain forbids one account from holding both. See [doc/account-model.md](../../doc/account-model.md). |
```
lotharking: Does `doc/account-model.md` exist?

packages/agent-sdk/src/utils/vtjscResolver.ts
```diff
@@ -0,0 +1,64 @@
...
+  const serviceId = `${ecosystem.did}#vpr-schemas-${schemaId}-vtjsc-vp`
+  const { didDocument } = await agent.dids.resolve(ecosystem.did)
+  const endpoint = didDocument?.service?.find(service => service.id === serviceId)?.serviceEndpoint
+  if (typeof endpoint !== 'string') {
+    throw new Error(
+      `Ecosystem ${ecosystem.did} publishes no VTJSC for schema ${schemaId} (no service ${serviceId})`,
+    )
+  }
+
+  const response = await fetch(endpoint)
```
lotharking: I’d prefer to use `fetchJson` with some typing to make the error cases easier to handle.

packages/agent-sdk/src/blockchain/VeranaChainService.ts
```diff
@@ -55,12 +55,20 @@ const {
...
 const PARTICIPANT_ROLE_ISSUER = 1
 
+// QueryListParticipantsRequest.response_max_size caps at 1024 and defaults to 64. The node applies
+// the other filters loosely, so ask for the largest page and match the fields again here.
+const PARTICIPANT_QUERY_MAX_SIZE = 1024
+
+// A simulation signs with an empty signature and runs against the state of the moment, so it
+// reports less gas than the delivery consumes. Cosmos SDK 0.47 made the difference larger (see
+// cosmos-sdk#16020), and cosmjs answers it with a default multiplier of 1.4. Not always enough.
+const DEFAULT_GAS_ADJUSTMENT = 1.5
```
lotharking: Why not handle this in the `constant` file and use it as required by the functions?

lotharking (approved):

PR #592, opened 2026-08-13
fix(vt-flow): validate schema before issuing credential

genaris:
Preventing a vt-flow from issuing a credential that does not have the proper claims set up.


tarunvadde (approved):
LGTM but do we really need this test file` packages/agent-sdk/tests/VtFlowOrchestratorClaimsValidation.test.ts`? If not then I think we can omit it.

PR #593, opened 2026-08-13
fix: self-tr integrity data calculation

genaris:
Quick fix for an issue found during devtest integration. Maybe we should just make something similar to the "digestJCS" instead of an SRI... Let's add this to the "TODO" list on the spec.


lotharking (approved):

lotharking (approved):
only a few comment

PR #594, opened 2026-08-18
fix: wait for the indexer subscribed acknowledgement before the catch-up

### LINKED ISSUE #574: Indexer event stream does not follow the connect-first order from spec v4-draft4

lotharking:
**Changes**
- The catch-up now waits for the `subscribed` message instead of the send callback.
- The 5s timeout stays and now logs a warning. Against an indexer that never
  acknowledges (older deployment) the agent degrades and keeps starting, rather than
  reconnecting in a loop.
- The e2e helper `IndexerSubscriber.connect` had the same race and now resolves on
  `subscribed` too.

### DIFF
--- a/packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
+++ b/packages/agent-sdk/src/blockchain/IndexerWebSocketService.ts
@@ -16,6 +16,7 @@ import {
   IndexerEventRecord,
   IndexerReadyMessage,
   IndexerSubscribeMessage,
+  IndexerSubscribedMessage,
   VeranaSyncState,
 } from './types'
 
@@ -46,7 +47,7 @@ export class IndexerWebSocketService {
   private generation = 0
   private chain: Promise<void> = Promise.resolve()
   private syncBuffer: IndexerEventRecord[] = []
-  private subscribeSent: (() => void) | null = null
+  private settleSubscribed: ((acknowledged: boolean) => void) | null = null
   private readonly indexer: VeranaIndexerService
   private readonly handlerRegistry: IndexerHandlerRegistry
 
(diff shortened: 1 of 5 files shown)


tarunvadde (approved):
One tiny optional thing: when the socket dies before the ack, the code still runs the full REST drain before the reconnect supersedes it. Distinguishing abandoned-by-close from timeout would skip that wasted drain.

PR #595, opened 2026-08-18
fix: resolve VPR schema references on any chain

### LINKED ISSUE #580: vpr references only resolve for two hardcoded chain ids
`mapToEcosystem` knows only `vna-testnet-1` and `vna-devnet-1` ([ecs.ts#L9-L12](https://github.com/verana-labs/vs-agent/blob/3a2b9528fedfe57fbbc975c775753ccd7b1eccac/packages/model/src/utils/ecs.ts#L9-L12)). An agent on any other chain cannot dereference a schema reference at all.

Related: the e2e stack still names its chain `vna-testnet-1` ([helpers.ts#L30](https://github.com/verana-labs/vs-agent/blob/3a2b9528fedfe57fbbc975c775753ccd7b1eccac/packages/agent-sdk/tests/e2e/helpers.ts#L30)), which maps to the public testnet indexer. Nothing calls `mapToEcosystem` in that suite today, so there is no leak yet, but the first test that exercises the JSC path without a precomputed digest would silently read a schema off the public host. The demo already moved to `vna-demo-1`.

Two parts: an env override for the indexer URL per chain, and renaming the e2e chain id.

lotharking:
**Changes:**
- mapToEcosystem resolves against a configurable registry instead of a hardcoded two-chain map, and throws on unresolvable VPR references instead of returning an unfetchable URI.
- The agent registers its own chain at startup from veranaChain.getChainId and VERANA_INDEXER_BASE_URL. No new environment variable.
- addDigestSRI falls back to the local schema when the request throws, not only when it returns a non-ok response.
- Removed the vna-testnet-1 fallback in stateMutations: without a chain connection the agent skips publishing instead of minting a JSC ref on the wrong chain.


tarunvadde (changes requested):
This one still references DEFAULT_CHAIN_ID after the merge with main, but the PR deleted the constant, so tsc fails with TS2304 and the package no longer builds. It came in with reconcileSelfIssuedEcsCredentials from the vs-account PR and git merged it clean since the hunks do not overlap. Since chain is already guaranteed non null right above, the fix that matches the rest of this PR is dropping the fallback, const chainId = chain.getChainId, with the same skip warn the other two call sites got.

genaris (approved):

PR #596, opened 2026-08-20
fix!: require a VERIFIED trust outcome in the VS-CONN-VS gate

### LINKED ISSUE #579: VS-CONN-VS accepts self-issued services: verified true with outcome NOT_TRUSTED passes the gate
Raised in https://github.com/verana-labs/vs-agent/pull/576 as a decision to make.

`assertVerifiableService` returns `verified` and only logs `outcome` ([assertVerifiableService.ts#L28](https://github.com/verana-labs/vs-agent/blob/3a2b9528fedfe57fbbc975c775753ccd7b1eccac/packages/agent-sdk/src/vtFlow/assertVerifiableService.ts#L28)), so a self-issued service with `verified: true` and `outcome: NOT_TRUSTED` passes. Anyone can run a stock vs-agent, let `setupSelfTr` self-sign its ECS credentials at startup, and pass every other agent's check. No chain registration, no participant, no fees, no validator.

Requiring a trusted outcome brings back the `vpr:` reference and the anchored digest, and `VERIFIED_TEST` needs a non-production registry. So this is a spec decision first: which outcomes may pass VS-CONN-VS, and where.

lotharking:
**Changes**
`resolveDID` returns `verified` and `outcome` as independent fields: a peer
that is a verifiable service but whose credentials come from no trusted
registry resolves as `verified: true` with `outcome: not-trusted`. The gate
returned `verified` alone, so those peers were accepted.

Gate on the outcome as well, accepting VERIFIED only. VERIFIED_TEST is
rejected as decided on the WG call.


genaris (commented):
What happens with the example https://github.com/verana-labs/vs-agent/blob/main/examples/vt-flow-demo/README.md after this code update? Is still working?

packages/agent-sdk/src/vtFlow/assertVerifiableService.ts
```diff
@@ -22,10 +22,13 @@ export function assertVerifiableService(
...
-      if (!verified) {
-        logger.warn(`[vt-flow] VS-CONN-VS rejected '${peerDid}': ${outcome} ${metadata?.errorMessage ?? ''}`)
+      const trusted = verified && outcome === TrustResolutionOutcome.VERIFIED
+      if (!trusted) {
+        logger.warn(
+          `[vt-flow] VS-CONN-VS rejected '${peerDid}': verified=${verified} outcome=${outcome} ${metadata?.errorMessage ?? ''}`,
+        )
       }
-      return verified
+      return trusted
```
genaris: I would like to make sure that we are compliant with https://github.com/verana-labs/verana-spec/blob/main/v4/vs-agent/spec.md#ecs-standalone-mode.

In particular, _"VS-1 MUST verify that Service-2 complies with [VS-REQ] and [VS-SVC], else VS-1 MUST NOT accept the connection, unless the purpose of the service provided by VS-1 is the issuance of [VT-ECS-ORG-CRED-W3C] or [VT-ECS-PERSONA-CRED-W3C] or [VT-ECS-SERVICE-CRED-W3C] credentials."_: when an Agent wants to perform the onboarding to get its Organization or Service credential, it does not have "real" credentials yet, so it will be unable to present credentials other than the self-signed ones. With current VS Agent code, will a validator VS Agent accept that request and proceed on the onboarding VT flow?

tarunvadde: Checked this against the spec text you quoted and it lines up. The new exemption allows exactly ORG, PERSONA and SERVICE, the same three the spec lists, and leaves UserAgent out. It also does not take the peer's word for it, the agent checks on chain that it is really the issuer or the named validator for that schema before letting the connection through. So a fresh agent with only self signed credentials can still onboard, and everyone else is now blocked, which was not the case before.

packages/agent-sdk/src/vtFlow/allowEcsIssuanceExemption.ts
```diff
@@ -0,0 +1,106 @@
...
+  options: AllowEcsIssuanceExemptionOptions,
+): VtFlowUnverifiedPeerExemptionHook {
+  const { indexer, trustedEcosystemDids } = options
+  return async ({ agentContext, peerDid, purpose }) => {
+    const logger = options.logger ?? agentContext.config.logger
+    const ownDid = options.ownDid()
+    if (!ownDid) return false
+    try {
+      const schemaId = purpose.participantId
+        ? await schemaOfPendingOnboarding(indexer, purpose.participantId, peerDid, ownDid)
```
tarunvadde: This is the first place where a value that comes straight from the peer's message goes into indexer.getParticipant, and that method puts the id into the URL as is. getParticipantSession right below it escapes with encodeURIComponent, this one does not. participantId is only checked as a string, so a peer can send something like 1/../../some/other/route and point the request at a different endpoint on our indexer. It cannot get past the gate since the reply has no participant field and we return false, but it does let a stranger pick URLs on our internal host. Either escape it or check it is an integer like the schemaId path already does.

packages/agent-sdk/tests/allowEcsIssuanceExemption.test.ts
```diff
@@ -0,0 +1,194 @@
...
+    trustedEcosystemDids: trustedEcosystemDids ?? [ECOSYSTEM_DID],
+    logger: logger as never,
+  })
+}
+
+function onboarding(participantId = '42') {
+  return { agentContext: {} as never, peerDid: PEER_DID, purpose: { participantId } }
+}
+
+describe('allowEcsIssuanceExemption: onboarding request', () => {
```
tarunvadde: The exemption has good tests but the actual fix has none. Every test here swaps assertVerifiableService for a fake one, and the older fake in apps/vs-agent/tests/vtFlow.test.ts still does the old broken check, verified only with no outcome. So if someone deletes the outcome check tomorrow nothing goes red. One test with verified true and outcome not-trusted that expects the connection to be refused would lock in the exact case from the issue.

tarunvadde (approved):
Approving, the gate change is right and it closes the case from issue #579. Left a few notes, the only one I would actually action soon is escaping the peer supplied participantId. Happy for that and the test to come as a follow up rather than holding this up.

genaris (changes requested):
The code might be doing its job, but it is introducing so many noise and I think we need to immediately simplify several things in `VsAgent` architecture. Maybe my comments would make this PR broader than it was initially, but I want to start reducing complexity rather than making it harder and leave all the cleaning work for a further PR: we always say that, and end up keeping legacy stuff forever.

apps/vs-agent/README.md
```diff
@@ -135,6 +135,12 @@ These variables connect the agent to the Verana network (permission management,
 
 > The agent maintains a persistent WebSocket connection to the indexer to receive updates about permissions, trust registries, and credential schemas. These events are used to keep the agent state in sync with the ledger.
 
+##### VS-CONN-VS trust gate
+
+Every vt-flow session is gated on trust resolution: the peer DID must resolve to a Verifiable Service whose trust chain reaches a production registry. A self-signed ECS credential is not enough.
+
+As allowed by [VS-CONN-VS], a Validator still accepts a peer that is not a Verifiable Service yet when the request is an ECS Organization, Persona or Service issuance — otherwise no agent could ever onboard. The exemption is not granted on the peer's word: for an onboarding request the peer must own a `PENDING` Participant entry naming this agent as its validator, and for a direct issuance request this agent must hold an active ISSUER Participant for the requested schema. In both cases the schema must be an ECS Organization, Persona or Service schema of an ecosystem listed in `TRUSTED_ECS_ECOSYSTEM_DIDS`. An Applicant never exempts its Validator: the peer it contacts must always be a Verifiable Service.
+
```
genaris: All of this explanation, if needed, should be written in verana-spec repo. We must try to always make it in sync with the behaviour of the implemented component (vs-agent in this case).

genaris (approved):
I think there are still severalt things to simplify, but they will come once all pending issues are implemented.

PR #618, opened 2026-08-20
fix: prevent duplicated connections by using publicDid tag

### LINKED ISSUE #617: Duplicate connection records stop all inbound DIDComm from a peer
An agent that both invites a peer and connects to the same peer keeps more than
one connection record for the same DID pair. `findByDids` accepts only one
record. It throws `RecordDuplicateError`. The error stops every inbound message
from that peer. The WebSocket transport hides the cause.

## Effect on devnet

`ecs-ecosystem` could not receive its Organization credential six times. The
validator (`ecs-org-issuer`) reached `CRED_OFFERED`. The applicant stayed at
`OR_SENT` with `credentialExchangeRecordId: null`. The indexer then reported
`trusted: false`, because the agent published no Organization credential.

`organization-vs` never had the failure. It never invites `ecs-org-issuer`, so
it keeps one connection record only.

## Mechanism

`DidCommConnectionRepository.findByDids` runs `findSingleByQuery` with this
query:

```js
{ $or: [
  { did: ourDid, theirDid },
  { did: ourDid, previousTheirDids: [theirDid] },
  { previousDids: [ourDid], theirDid },
]}
```

`Repository.findSingleByQuery` throws `RecordDuplicateError` when the query
matches more than one record.

`ecs-ecosystem` held these two records for the same peer:

| record | `did` | `theirDid` | `previousDids` | matched by |
|---|---|---|---|---|
| `2ea82b91` | `did:peer:4zQmSckbU…` | issuer webvh | `[ecs-eco webvh]` | clause 3 |
| `111eec7e` | ecs-eco webvh | issuer webvh | `[]` | clause 1 |

The agent creates the first record when the peer answers its **reusable** public
OOB invitation. `DidCommMessageReceiver.rotateInviterDidForV2OOB` then gives the
agent a fresh peer DID and moves the public DID into `previousDids`.

The agent creates the second record when it starts its own onboarding process
against the same peer.

From that point, `findConnection` throws on every inbound message from that
peer.

## The hidden error

`VsAgentWsInboundTransport.listenOnWebSocketMessages` logs the message text only:

```ts
} catch (error) {
  this.logger.error('Error processing message')
}
```

`HttpInboundTransport` logs `error.message`. Please make the WebSocket transport
do the same. The devnet log gave no cause for six attempts.

## Proof

1. Delete every connection record between the two agents.
2. Start a new onboarding process.
3. Delete the rotated `did:peer` record after the applicant sends the onboarding
   request, and before the validator offers the credential.
4. The flow goes to `COMPLETED`.

Participant 90 of `ecs-ecosystem` received its Organization credential with this
sequence. The indexer now answers `trusted: true` with both
`ServiceCredential` and `OrganizationCredential`.

## Proposed corrections

1. **Log the cause.** `VsAgentWsInboundTransport` must log `error.message` and
   the error object, as `HttpInboundTransport` does. This is a one-line change
   and it makes every future failure of this class visible.
2. **Do not create a second connection.** Before the agent starts an onboarding
   process, it must look for a connection record that already holds the pair
   (`our public DID`, `peer public DID`) and use it.
3. **Tolerate duplicates on inbound.** `findConnection` must catch
   `RecordDuplicateError` from `findByDids` and choose the newest ready record,
   in the same way it already does for `findByTheirDid`.

Correction 1 is necessary. Correction 3 gives the strongest protection, because
it also covers records that other flows create.

genaris:
Addresses https://github.com/verana-labs/vs-agent/issues/617 by using a `publicDid` tag for each DidCommConnection. This query is quite fast, since it will only use a tag. 

Additionally, it will delete the connection record automatically in case of any issue (e.g. timeout). This will make it easier to prevent stale connection records.

### DIFF
--- a/packages/agent-sdk/src/utils/agent.ts
+++ b/packages/agent-sdk/src/utils/agent.ts
@@ -1,7 +1,12 @@
 import type { VsAgent } from '../agent/VsAgent'
 
 import { parseDid } from '@credo-ts/core'
-import { DidCommHandshakeProtocol, DidCommMessage, type DidCommVersion } from '@credo-ts/didcomm'
+import {
+  DidCommConnectionRepository,
+  DidCommHandshakeProtocol,
+  DidCommMessage,
+  type DidCommVersion,
+} from '@credo-ts/didcomm'
 
 /**
  * Creates an out of band invitation that will equal to the public DID in case the agent has one defined,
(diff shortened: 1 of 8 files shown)


packages/agent-sdk/src/utils/agent.ts
```diff
@@ -59,6 +64,55 @@ export async function createInvitation(options: {
...
+    did: peerPublicDid,
+    ourDid: agent.did,
+    label: agent.label,
+    didCommVersion: 'v2',
+  })
+  if (!connectionRecord) throw new Error(`Failed to establish a DIDComm connection to ${peerPublicDid}`)
+
+  try {
+    // Tag the record with the public DID of the peer, to find it on the next call.
+    connectionRecord.setTag('publicDid', peerPublicDid)
```
lotharking: This tag is only valid for a new wallet. This error will persist until the wallet is restarted.

lotharking (approved):

tarunvadde (approved):
LGTM.

PR #621, opened 2026-08-20
feat: admin api add a v2 api router

### LINKED ISSUE #597: [Admin API] Add a v2 API router
Add `version: '2'` controllers under a new `src/controllers/admin/v2/` tree, one directory per
scope: `auth`, `agent`, `didcomm`, `openid4vc`, `anoncreds`, `vt`. URI versioning is already
enabled in `setupAgent.ts`, so both versions can serve at once.

v1 keeps working through Phases 1–6 and is removed in Phase 7 (R2).

**Files:** `apps/vs-agent/src/utils/setupAgent.ts`, `apps/vs-agent/src/controllers/admin/`

**Expected result:** an empty v2 controller per scope answers `404` on unknown paths and appears in Swagger under a v2 tag. No v1 behaviour changes.

lotharking:


genaris (approved):
I think I'd prefer to use an "V2" prefix rather than adding it in the middle of the class name (e.g. `V2AnoncredsController.ts`). Also we can rename existing controllers to also follow this rule.

PR #624, opened 2026-08-21
feat: admin api v2 liveness and readiness probes

### LINKED ISSUE #603: [Admin API] Implement VSA-ADM-AG-LIVE and VSA-ADM-AG-READY
`GET /v2/agent/health/live`: answers `200` as soon as the HTTP listener accepts a connection, and before the bootstrap sequence completes. It MUST NOT fail because an external dependency failed: not the Verana RPC endpoint, not the indexer, not the peer of a flow. Fail it only for a condition a restart repairs.

`GET /v2/agent/health/ready`:  `200` when ready; `503` with code `NOT_READY` otherwise, the message naming the pending bootstrap step. Not ready until every bootstrap step completes, and again while catching up with the indexer after a reconnection: until catch-up finishes the agent's view of on-chain state is stale.

Mark both probes `@AccessMode('PUBLIC')` so they serve in both `ADMIN_API_AUTH_MODE` values, to trusted-network and external callers alike, with no bearer token and no allowlist check. Neither may ever answer `401` or `403`.

Because anything reaching the port reaches both probes, assert that no secret, token, Verana account address, DID, or peer identifier appears in a probe body.

lotharking:
**Changes**
- Add GET /v2/agent/health/live and /v2/agent/health/ready, both public so neither can answer 401 or 403.
- New BootstrapState gating readiness on the agent's own bootstrap steps and on the indexer catch-up, with the ECS bootstrap recorded but not gating.
- Expose syncStatus on IndexerWebSocketService; an unreachable indexer keeps the agent ready, a catch-up does not.
- Add the NOT_READY code so the v2 error envelope preserves it, and log an envelope only when its code is INTERNAL.
- Chart: startupProbe/livenessProbe on live, readinessProbe on ready.

### DIFF
--- a/apps/vs-agent/src/admin.module.ts
+++ b/apps/vs-agent/src/admin.module.ts
@@ -24,6 +24,7 @@ import {
   V1VsAgentController,
   MESSAGE_HANDLERS,
 } from './controllers'
+import { BOOTSTRAP_STATE, BootstrapState } from './common'
 import { AdminAuthGuard, AdminAuthService, V1AuthController } from './security'
 import { UrlShorteningService } from './services/UrlShorteningService'
 import { VsAgentService } from './services/VsAgentService'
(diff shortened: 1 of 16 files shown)


genaris (approved):

tarunvadde (approved):
LGTM.

Issue #623, opened 2026-08-21
remove backward compatibility


lotharking:
Hey @genaris, if we don’t need to maintain backward compatibility, we must remove the legacy code related to this.

_Originally posted by @lotharking in https://github.com/verana-labs/vs-agent/issues/618#issuecomment-5373230691_

genaris:
I'm closing this issue because it is so broad and it generates confusions. We'll open specific issues for each configuration parameter or API method we should remove.

PR #626, opened 2026-08-24
feat(vs-agent)!: admin API v2 vt flows and service endpoints

### LINKED ISSUE #614: [Admin API] Create VT endpoints for V2
In V2, most of the /vt/ scoped endpoints are similar to the ones already working for v1, so this single issue englobes the work of adding them for v2, as specified in **VSA-ADM-VT-FL** (Flow Management) and **VSA-ADM-VT-SE** (Service Endpoint Management).

Note that existing methods for V1 (linked-credentials and json-schema-credentials) will still exist, but only GET method will be available, since POST and DELETE are incompatible with current specified VS Agent behaviour.

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/614. 

Adds the /v2/vt scope per VSA-ADM-VT-FL and VSA-ADM-VT-SE from the merged V2 API spec (https://github.com/verana-labs/verana-spec/pull/55), flows and service endpoints with camelCase DTOs, the v2 error envelope (409 INVALID_STATE, 404 UNKNOWN_ID) and keyset pagination on the list methods. The pagination module is a local stand-in until https://github.com/verana-labs/vs-agent/issues/599 lands, both list methods should move onto it then.

The v1 vt credential routes are removed entirely rather than trimmed to GET, per the direction in https://github.com/verana-labs/vs-agent/issues/623. The standalone v1 issue and revoke POSTs stay for that same issue since v2 keeps registry revocation in a different scope. Route shapes follow the spec as of https://github.com/verana-labs/verana-spec/pull/58, serviceEndpointId path params included.

Note for review timing, this touches VtFlowsService which https://github.com/verana-labs/vs-agent/pull/625 also reworks, whichever lands second rebases.


apps/vs-agent/src/controllers/admin/verifiable/V1TrustController.ts
```diff
@@ -132,290 +125,4 @@ export class V1TrustController {
       anoncredsRevocationRegistryIndex,
     })
   }
-
-  @Get('linked-credentials')
```
lotharking: > Note that existing methods for V1 (linked-credentials and json-schema-credentials) will still exist, but only GET method will be available, 

The GET method should still be persisted as proposed in the issue, but it seems to have been removed here.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -98,6 +133,34 @@ export class VtFlowsService {
...
+        vtFlowRecordId: record.id,
+        state: VtCredentialState.Revoked,
+        reason,
+      })
+    })
+  }
+
+  private async revokeIssuedCredential(agent: VsAgent, record: VtFlowRecord): Promise<void> {
+    record.assertState([VtFlowState.Completed, VtFlowState.CredRevoked])
+    if (!record.credentialExchangeRecordId) return
```
lotharking: Should this also leave `UNSUPPORTED_FORMAT` aside?

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
```diff
@@ -98,6 +133,34 @@ export class VtFlowsService {
...
+    const credential = await agent.didcomm.credentials.findById(record.credentialExchangeRecordId)
+    if (!credential) return
+    const registryId = credential.getTag('anonCredsRevocationRegistryId')
+    const revocationId = credential.getTag('anonCredsCredentialRevocationId')
+    if (typeof registryId !== 'string' || !revocationId) {
+      throw new AdminApiError(
```
lotharking: 'UNSUPPORTED_FORMAT' is a bare string outside AdminApiErrorCode. Should method-specific codes live in the enum too?

apps/vs-agent/src/controllers/admin/v2/vt/V2VtFlowsController.ts
```diff
@@ -0,0 +1,113 @@
...
+  @ApiParam({ name: 'participantSessionId', type: String })
+  @ApiOkResponse({ type: VtFlowRecordDto })
+  @ApiNotFoundResponse()
+  @ApiConflictResponse()
+  public validateFlow(@Param('participantSessionId') participantSessionId: string): Promise<VtFlowRecordDto> {
+    return this.service.validateAndOfferCredential(participantSessionId)
+  }
+
+  @Post(':participantSessionId/revoke-credential')
+  @AccessMode('CORPORATION', ['/verana.pp.v1.MsgRevokeParticipant'])
```
lotharking: MsgRevokeParticipant never appears in a VsOperatorAuthorization, so callerHoldsVsOperatorGrant always returns false and this route 403s for every external caller. Is there a msgType that actually covers it?

apps/vs-agent/src/controllers/admin/v2/vt/V2VtFlowsController.ts
```diff
@@ -0,0 +1,113 @@
...
+} from '../../vt-flow/dto/flow-requests.dto'
+import { VtFlowRecordDto, VtFlowRecordPageDto } from '../../vt-flow/dto/vt-flow-record.dto'
+
+@ApiTags('v2/vt')
+@Controller({ path: 'vt/flows', version: '2' })
+export class V2VtFlowsController {
+  public constructor(@Inject(VtFlowsService) private readonly service: VtFlowsService) {}
+
+  @Get()
+  @AccessMode('CORPORATION', [
```
lotharking: The other v2 controllers declare @AccessMode('INTERNAL') at class level, and the spec says Admin API methods declare no msgType. Should v2 flows follow the same convention?

apps/vs-agent/src/controllers/admin/verifiable/V1TrustController.ts
tarunvadde: That was deliberate and it is in the PR description, the routes went entirely rather than GET only, following #623 which asks to remove backward compatibility in the same change. If we want the issue text to win over #623 here I am happy to restore the two GETs, @genaris wdyt?

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
tarunvadde: Both no exchange record and a vanished exchange now answer UNSUPPORTED_FORMAT instead of pretending the revocation happened.

apps/vs-agent/src/controllers/admin/vt-flow/VtFlowsService.ts
tarunvadde: Agreed, added UnsupportedFormat to AdminApiErrorCode and the service uses the enum member now.

apps/vs-agent/src/controllers/admin/v2/vt/V2VtFlowsController.ts
tarunvadde: Thanks, idk how I missed it. Gone now, both v2 vt controllers declare INTERNAL at class level with no msgTypes, matching the other v2 controllers and the spec.

apps/vs-agent/src/controllers/admin/v2/vt/V2VtFlowsController.ts
tarunvadde: Done

apps/vs-agent/src/controllers/admin/verifiable/V1TrustController.ts
genaris: I think #614 is a bit better explained than #623, so we can keep the GET methods for the moment. But honestly, it is not critical so if that represents adding noise to current code just to make it compatible with V1 and V2 APIs, I prefer to just drop V1 methods just from the beginning.

Keep in mind that our expectation for "final" v2.0.0 is to drop V1 API entirely: we are just keeping it to make the transition easier these days, when we are working on the testing of verana demos on devnet. But I prefer to have some more noise in those demos rather than in this source code.

tarunvadde: Restored the two GETs as they were.

lotharking (approved):

genaris:
@tarunvadde can you fix the conflicts and merge this?

PR #627, opened 2026-08-24
feat(vs-agent)!: remove AGENT_DIDCOMM_VERSIONS

### LINKED ISSUE #607: Drop AGENT_DIDCOMM_VERSIONS
Remove `AGENT_DIDCOMM_VERSIONS`. Supporting both envelopes is a conformance requirement, not a deployment option. In related endpoints, default an omitted `didcommVersion` to v2.

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/607. 

Supporting both DIDComm envelope versions is a conformance requirement, not a deployment option, so the env var is gone and the agent always enables v1 and v2. The versions are passed explicitly since the underlying defaults are v1 only and removal by omission would silently drop v2. Invitation endpoints default an omitted didcommVersion to v2.

Breaking for deployments that set the var, it is ignored now, and a v2 only deployment starts publishing the v1 did-communication service again, which is the intended conformance behavior.


packages/agent-sdk/src/utils/agent.ts
lotharking: What about the `setupBaseDidComm` file? Why does this preserve the default V1 configuration?

tarunvadde: Thanks. Fixed.

lotharking (approved):

PR #629, opened 2026-08-24
feat: admin api implement vsa adm ag info

### LINKED ISSUE #602: [Admin API] Implement VSA-ADM-AG-INFO
/v2/agent/info

lotharking:
Changes

getAgentInfo serves GET /v2/agent/info, returning the DID the agent
created on its first startup and the running application version. The
agent still starts without a DID today, so the field is omitted when
there is none instead of failing the request.

Admin API v2 authorizes an external caller with the allowlist alone,
without VPR grants, but the guard rejected a CORPORATION route that
declares no message type. It now serves one once the account passes
the allowlist. The v1 flow methods are unchanged, since all of them
declare message types.


tarunvadde (approved):
Looks good. One note for later, once startup DID creation lands the did field stops being optional, worth tightening the DTO then.

genaris (approved):

PR #632, opened 2026-08-24
fix: withdraw the self-issued ECS credential when its ISSUER participant is revoked

### LINKED ISSUE #630: Remove a self-issued credential when its issuer participant becomes revoked.
Extend the RevokeParticipant handler. Today it looks only for a HOLDER
participant of this agent. It must also look for an ISSUER participant of this
agent, and remove each stored ECS credential that anchors against it.

lotharking:
An ECS credential anchored against a revoked ISSUER makes the whole DID fail VS-CONN-VS,
and nothing removed it: `removeHolderTrustCredentialIfRevoked` returns for any non-HOLDER role.

## Changes

- `RevokeParticipant` / `SlashParticipantTrustDeposit` withdraw the credentials anchored
  against this agent's ISSUER participant and republish the self-issued default.
- Same withdrawal at startup for already revoked or slashed participants, which no event reaches.
- `rebindEcsCredentialSchema` persists `issuerParticipantId`; existing wallets are backfilled
  by the next reconciliation.
- `removeTrustCredential` looked entries up differently from `deleteMetadataEntry` and left the
  DID Document service dangling for self-issued entries.


packages/agent-sdk/src/utils/trustCredentialStore.ts
```diff
@@ -465,8 +477,44 @@ export async function rebindEcsCredentialSchema(
...
+  publicApiBaseUrl: string,
+  issuerParticipantId: number,
+  defaults: SelfTrDefaults,
+): Promise<string[]> {
+  const didRecord = await getDidRecord(agent)
+  const withdrawn: string[] = []
+
+  for (const [jscUrl, value] of Object.entries(didRecord.metadata.get('_vt/vtc') ?? {})) {
+    const entry = value as { issuerParticipantId?: number; schemaKey?: string }
+    if (entry.issuerParticipantId !== issuerParticipantId) continue
```
tarunvadde: One gap, and whether it matters or not I'm not sure. Entries written by the current dev releases have no issuerParticipantId, only the rebind loop stamps it and that loop skips revoked issuers. So for a wallet whose issuer was already revoked before this upgrade the startup catch up matches nothing and the stale credential stays published. Wallets created after the upgrade are fine, they get stamped while the issuer is alive.

If wallets are disposable across upgrades like the rest of devnet this is a non issue, then the startup loop only serves future offline revocations and maybe the description should say so. If any wallet survives the upgrade it needs a fallback for unstamped entries or a one time stamp migration.

tarunvadde (changes requested):

packages/agent-sdk/src/utils/trustCredentialStore.ts
genaris: That was actually what I was referring to when saying "no backwards compatibility required". Don't worry about anything done after v1.12.0 release, especially if it is about VT concepts.

genaris (approved):

tarunvadde (approved):

PR #633, opened 2026-08-25
fix: withdraw VTJSCs of ecosystems the agent no longer controls

### LINKED ISSUE #631: Remove a VTJSC when the agent stops controlling its Ecosystem
In `reconcileVtjscPublications`, collect the schema references the pass reconciled, then remove each `_vt/jsc` entry, and each DID Document service, that the set does not contain.

lotharking:
## Changes

- `withdrawVtjscPublications` (`trustCredentialStore.ts`): drops the given `_vt/jsc`
  entries and their linked VPs, publishing the document once.
- `withdrawUncontrolledVtjscPublications` (`stateMutations.ts`): runs after the
  publication pass and withdraws entries whose Ecosystem is archived or belongs to
  another Corporation.
- The publication pass now skips archived Ecosystems.


genaris (commented):
Can you check for conformance with current VT and VPR specifications? IIRC, VTCJSC for archived ecosystem should remain present, because otherwise credentials based on them wouldn't be verifiable anymore. However I'm not sure for the case where an Agent loses control of an ecosystem (!).

lotharking:
Hey @genaris , Archived is no longer a criterion, and detaching now keeps the entry stored so the VTJSC URL that issued the credentials referenced by `credentialSchema.id` keeps resolving. When control is lost, we detach but keep serving for the same verifiability reason. One caveat is that control is detected through `corporation_id`, which is the signal the publication pass already used before this PR, rather than `Ecosystem.did` as defined by [VSA-VTI-VTJSC]. Since `corporation_id` is immutable on-chain, an Ecosystem handed to a sibling agent of the same Corporation through `UpdateEcosystem` would not be detected. Aligning both passes on `Ecosystem.did` would be a separate change.

tarunvadde (approved):
LGTM.

PR #634, opened 2026-08-25
feat(vs-agent)!: derive, create and publish the public DID at startup

### LINKED ISSUE #600: Derive the DID from PUBLIC_API_BASE_URL
Remove `AGENT_PUBLIC_DID`. An operator cannot know a `did:webvh` DID in advance, because its self-certifying identifier exists only after creation: so the DID is an output of the agent, not an input to it.

- Make `PUBLIC_API_BASE_URL` required. Reject a URL carrying a username or a password.
- Add `AGENT_PUBLIC_DID_METHOD`: `webvh` (default) or `web`. Reject any other value at startup.
- Implement the location derivation: host, percent-encoded port, colon-separated path segments. Here it is important to determine where the did.json/did.jsonl will reside (i.e. no .well-known when it's under a path of a domain).

**Files:** `src/config/constants.ts`, `src/main.ts` (lines 61, 186–189)

**Expected result:** table-driven tests cover the three location cases in the spec, and startup fails with a named error on a bad method value.

> **Note**
> If it becomes difficult to implement subpath creation on did:webvh due ot the underlying libraries, we can defer this support to a further version. Raise a follow-up issue in such case.

tarunvadde:
Closes https://github.com/verana-labs/vs-agent/issues/600 and https://github.com/verana-labs/vs-agent/issues/601.

AGENT_PUBLIC_DID is gone. A did:webvh SCID only exists after creation, so the DID is now an output of the agent. PUBLIC_API_BASE_URL is required and rejects userinfo, AGENT_PUBLIC_DID_METHOD picks webvh (default) or web, and the location comes from the URL, host, %3A encoded port, colon joined path segments.

On a cold start the agent creates the keys, composes the DID, publishes it and persists it before DIDComm starts accepting messages. On later starts it loads the persisted DID and refuses to boot if the location or the method changed, naming both values, so it never mints a second DID and never discards the credentials bound to the first.

The document is served only on the shape the location implies, under /.well-known for a bare domain and outside it for a path, since serving both would let two agents on the same domain claim the same well known URL. A path deployment assumes the proxy strips the base path.

Breaking, AGENT_PUBLIC_DID is removed and PUBLIC_API_BASE_URL becomes required. Changing either the base URL or the method on an existing wallet now fails the boot on purpose.


packages/agent-sdk/src/agent/VsAgent.ts
```diff
@@ -177,7 +203,7 @@ export class VsAgent<TModules extends BaseAgentModules = BaseAgentModules> exten
 
           await this.dids.create({
```
lotharking: Why don't we check whether an error occurred during the saving process?

genaris (approved):
LGTM, I left just some nitpicks here.

PR #638, opened 2026-08-26
fix: rebuild the self-issued ECS credential when the stored one no longer matches the DID Document

### LINKED ISSUE #616: An agent cannot drop or rebuild a self-issued ECS credential that no longer verifies
This issue describes a problem I had during the integration of verana-demos with devnet, where I needed to create different ecosystems, issue and revoke VTCs, etc. They probably don't reflect the most typical use case, but the topics written here might be valuable to make the whole implementation (and the specs) consistent and supporting some edge, recovery cases.

## Summary

A VS Agent publishes its ECS credentials as Linked Verifiable Presentations in its
DID Document. When one of these credentials becomes invalid, the agent keeps it and
publishes it. The agent has no code path that removes it, and no code path that
builds it again.

One invalid credential makes the whole DID fail the VS-CONN-VS check. The agent then
cannot start or accept a vt-flow exchange, so it cannot obtain the credential that
would replace the invalid one. The deployment is stuck.

We found this on devnet with two agents. See "How to reproduce".

## What the agent does today

Two mechanisms write these credentials:

- `setupSelfTr` builds a self-issued credential for each ECS schema at start.
- `reconcileSelfIssuedEcsCredentials` (`stateMutations.ts`) binds that credential to
  the Ecosystem's VTJSC and anchors its digest, for each ISSUER participant the agent
  holds.

Two mechanisms remove a credential, and neither one applies here:

- `removeHolderTrustCredentialIfRevoked` (`stateMutations.ts`) removes a **received**
  credential when the agent's own **HOLDER** participant becomes revoked. It returns
  immediately for any other role, and it ignores self-issued credentials.
- `DELETE /v1/vt/linked-credentials` removes a credential through the admin API. The
  `IndexerDisabledGuard` rejects this request when `VERANA_INDEXER_BASE_URL` is set,
  which is true for every deployment that uses the VPR.

`generateVerifiablePresentation` (`setupSelfTr.ts`) decides to build the credential
again from `integrityData`, which is a digest of the claims, the type and the schema
reference. A stored presentation that no longer verifies has the same `integrityData`
as before, so the agent keeps it.

### Case 1 — the issuer participant becomes revoked

1. Let an agent self-issue its ECS Organization credential. The agent anchors the
   digest against its own ISSUER participant.
2. Revoke that ISSUER participant.
3. Resolve the agent's DID.

The resolver reports:

```
Participant for the specified DID: did:webvh:...:ecs-ecosystem.devnet.verana.network
for schema 1 is REVOKED
```

The agent receives the `RevokeParticipant` event. It removes nothing, because
`removeHolderTrustCredentialIfRevoked` accepts only the HOLDER role.

### Case 2 — the credential points to the wrong VTJSC

1. Let an agent bind its ECS Service credential to a VTJSC that is not the one the
   ECS Ecosystem publishes. An earlier release did this, because it bound a credential
   for every schema of the agent's own Corporation.
2. Resolve the agent's DID.

The resolver reports:

```
The verifiable credential proof is not valid with: One or more verifiable
credentials failed signature verification.
```

A restart does not correct the binding. A change of the claims makes the agent build
the credential again, but the agent writes it to the same stale key, so the DID
Document continues to present the stale binding.

### Case 3 — a VTJSC of an Ecosystem the agent no longer controls

1. Start an agent bound to a Corporation that controls an Ecosystem. It publishes a
   VTJSC for each schema of that Ecosystem.
2. Bind the same agent to another Corporation, or archive the Ecosystem.
3. Read the DID Document.

The VTJSCs of the first Ecosystem are still there, next to the new ones. On devnet an
agent presented eight, for two Ecosystems, while it controlled one.

`reconcileVtjscPublications` only adds. It walks the Ecosystems of the agent's current
Corporation and creates or refreshes one VTJSC for each schema. Nothing compares that
set against what the DID Document already presents, so an entry that no longer belongs
stays for the life of the wallet. `updateVtcEntries`, the one cleanup that exists, is
gated on `key === '_vt/vtc'` and never touches a `_vt/jsc` entry.

A trust credential does not have this problem: `rebindEcsCredentialSchema` removes the
entries that carry the same presentation URL before it writes the new one.

Nothing consumes the stale copies, because a resolver finds the publisher through
`Ecosystem.did` and that DID no longer points at this agent. The DID Document
nevertheless claims that the agent publishes the schemas of an Ecosystem it does not
control, which misleads a reader and any audit of what the service governs.

## Why this blocks the deployment

VS-CONN-VS is symmetric. The applicant checks the validator, and the validator checks
the applicant. An agent with one invalid credential fails both directions. It cannot
receive a credential over the onboarding process, which is the only supported way to
replace the invalid one.

The operator has no way out either. The admin API is closed, and the participant
lifecycle messages do not touch the DID Document.

## Proposed correction

Some ideas. To be carefully analyzed for conformance with Verifiable Trust and Verifiable Public Registry specs:

1. **Remove a self-issued credential when its issuer participant becomes revoked.**
   Extend the `RevokeParticipant` handler. Today it looks only for a HOLDER
   participant of this agent. It must also look for an ISSUER participant of this
   agent, and remove each stored ECS credential that anchors against it.

2. **Remove a VTJSC when the agent stops controlling its Ecosystem.** In
   `reconcileVtjscPublications`, collect the schema references the pass reconciled, then
   remove each `_vt/jsc` entry, and each DID Document service, that the set does not
   contain.

3. **Verify the stored presentation before the agent keeps it.** In
   `generateVerifiablePresentation`, `integrityData` is not sufficient. Add a check of
   the stored presentation. Build the credential again when one of these is true:
   - the signature of the stored presentation does not verify;
   - `credentialSchema.id` is not the VTJSC that `resolveJsonSchemaCredentialId`
     returns for the schema.

   The second condition also corrects the stale bindings that earlier releases wrote.

lotharking:
**Changes**
- `generateVerifiablePresentation` now revalidates the stored credential before the cache
  hit: `issuer`, `credentialSchema.id` against the record key, and every
  `proof.verificationMethod` against the DID Document's current `assertionMethod`. Any
  mismatch falls through to the rebuild path that already existed. The verification method
  id is derived from the key material, so a key rotation always changes it.
- The self-issued default is stored detached while another entry already serves the same
  presentation URL. `setupSelfTr` runs before `reconcileVtjscPublications`, so every boot
  re-inserted a competing entry for the URL a rebound credential owns.
- `findMetadataEntry` stops depending on insertion order: among entries matching one URL,
  the one the DID Document announces wins, with a detached entry as the fallback.

### DIFF
--- a/apps/vs-agent/tests/e2e/fullLifecycle.e2e.test.ts
+++ b/apps/vs-agent/tests/e2e/fullLifecycle.e2e.test.ts
@@ -621,6 +621,57 @@ describe('v4 full lifecycle on a live chain and indexer', () => {
   )
 
   it(
+    'rebuilds the self-issued ECS credential when its stored proof leaves the DID Document',
+    async () => {
+      const baseUrl = applicant.publicApiBaseUrl
+      const didRecordOf = async () => (await applicant.dids.getCreatedDids({ did: applicant.did }))[0]
+      const storedEntry = async (key: string) => (await didRecordOf()).metadata.get('_vt/vtc')?.[key]
+
+      const jscUrl = await resolveJsonSchemaCredentialId(
+        applicant,
+        indexer,
+        serviceSchemaId,
+        seederChain.getChainId,
+      )
+      const rebind = async () =>
+        await rebindEcsCredentialSchema(
+          applicant,
+          baseUrl,
+          String(serviceSchemaId),
+          'ecs-service',
+          selfTrDefaults,
+          jscUrl,
+          applicantIssuerParticipantId,
+        )
+
+      await rebind()
+      const stored = await storedEntry(jscUrl)
+      expect(stored?.credential?.proof?.verificationMethod).toBeDefined()
+      const integrityData = stored.integrityData
+
+      const rotated = `${applicant.did}#rotated-key`
+      const didRecord = await didRecordOf()
+      const record = didRecord.metadata.get('_vt/vtc') ?? {}
+      record[jscUrl].credential.proof.verificationMethod = rotated
+      didRecord.metadata.set('_vt/vtc', record)
+      await applicant.context.dependencyManager.resolve(DidRepository).update(applicant.context, didRecord)
+      expect((await storedEntry(jscUrl)).credential.proof.verificationMethod).toBe(rotated)
+
+      await rebind()
+
+      const rebuilt = await storedEntry(jscUrl)
+      const assertionMethods = ((await didRecordOf()).didDocument?.assertionMethod ?? []).map(entry =>
+        typeof entry === 'string' ? entry : entry.id,
+      )
+      expect(rebuilt.integrityData).toBe(integrityData)
+      expect(rebuilt.credential.proof.verificationMethod).not.toBe(rotated)
+      expect(assertionMethods).toContain(rebuilt.credential.proof.verificationMethod)
+      expect(rebuilt.credential.credentialSchema.id).toBe(jscUrl)
+    },
+    SETUP_TIMEOUT_MS,
+  )
+
+  it(
     'withdraws the self-issued ECS credential when its ISSUER participant is revoked on chain',
     async () => {
       const baseUrl = applicant.publicApiBaseUrl
--- a/packages/agent-sdk/src/utils/setupSelfTr.ts
+++ b/packages/agent-sdk/src/utils/setupSelfTr.ts
@@ -314,6 +314,43 @@ export async function signerW3c(
   }
 }
 
+interface StoredSelfIssuedCredential {
+  issuer?: string | { id?: string }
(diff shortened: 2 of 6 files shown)


tarunvadde (approved):
LGTM.

tarunvadde:
Worth knowing that generateVerifiableCredential still has the same key blind cache gate you fixed here in generateVerifiablePresentation. Not from this PR and not blocking, but it is the obvious follow up.

lotharking:
> Worth knowing that generateVerifiableCredential still has the same key blind cache gate you fixed here in generateVerifiablePresentation. Not from this PR and not blocking, but it is the obvious follow up.

I’m not sure if we need to update the `generateVerifiableCredential` function, since it is currently only used by the SelfTr implementation. In the last WG call, it was mentioned that SelfTr should be removed, so this may not be necessary.

PR #639, opened 2026-08-26
feat: admin api didcomm connection methods

### LINKED ISSUE #608: [Admin API] DIDComm connection methods
Implement VSA-ADM-DC-CN methods for v2 API:

- GET `/v2/didcomm/connections`
- GET `/v2/didcomm/connections/{connectionId}`
- DELETE `/v2/didcomm/connections/{connectionId}`

lotharking:
### Changes
- **`listConnections`** accepts the nine filters the spec names — `outOfBandId`,
  `state`, `role`, `did`, `theirDid`, `threadId`, `invitationDid`,
  `didcommVersion`, `mediatorId` — plus the shared `limit` / `cursor`. It answers
  `{ items, nextCursor }` with a keyset cursor ordered on `createdAt|id`. The
  cursor is bound to the method and filter set that minted it, so a replay against
  another filter set is refused with `INVALID_CURSOR`.
- **`getConnection`** returns the record through the same mapper as the listing,
  so both answer the identical shape. This is where v1 diverges: it maps by hand
  in `list` and returns `record.toJSON()` in `get`, yielding two different shapes.
- **`deleteConnection`** answers `204` with an empty body, and reports an unknown
  id as `UNKNOWN_ID` rather than letting `RecordNotFoundError` collapse into
  `INTERNAL`. Credo's `deleteById` also drops the routing the connection held on
  its mediator, which the spec allows as a `MAY`.


tarunvadde (approved):
One thing I checked and want to note so nobody rediscovers it. Filtering with didcommVersion=v1 returns an empty page, because credo only ever writes v2 on the record. That is the same on GET /v1/connections today, so it is not from this PR, but the v2 swagger enum now advertises a value that can never match.

